PR #31, opened 2018-05-29
Make setpoints vary subhourly.

nealkruis:


chipbarnaby:
Structural changes look good.

Remaining issue is whether there are any runtime values that are derived from the setpoints at intervals longer than subhour.  All uses should be reviewed to make sure they are in subhour code paths.

PR #35, opened 2018-06-04
Make special script for debugging C Preprocessor issues.

nealkruis:
@chipbarnaby this doesn't exactly solve our problem, but it helps debug the problems in the future. I suspect there was something strange happening with CMake where it wasn't checking to make sure dependencies existed and would skip the command entirely if there weren't any existing dependencies.

### DIFF
--- a/cmake/preprocess.cmake
+++ b/cmake/preprocess.cmake
@@ -1,0 +1,9 @@
+execute_process(
+  COMMAND "${compiler}" -c -EP -nologo -I "${source_dir}" -Tc "${source_dir}/${file}.def"
+  OUTPUT_FILE "${out_dir}/${file}.i"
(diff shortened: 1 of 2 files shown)


nealkruis:
The commands executed by add_custom_command cannot do anything with the return code from its command. By replace the command calling the preprocessor with a script that calls the preprocessor we can investigate why the preprocessor fails and halt the build process explicitly.

chipbarnaby:
Awkward but one does what one needs to do.

--
Charles S. "Chip" Barnaby, FIBPSA / FASHRAE
chipbarnaby@gmail.com
879 Bean Rd, Moultonborough, NH 03254  USA
(o) +1 603 253 4593 or +1 603 253 4400   (m) +1 781 883 4593


----- Original message -----

PR #36, opened 2018-06-18
Add Kiva ground heat transfer calculations

nealkruis:


nealkruis:
@chipbarnaby I have some input related questions for you:

- TOP soilDiff: We now need conductivity, density, and specific heat for Kiva. Do we still want to have this input if we can calculate it?
- SURFACE sfDepthBG: Kiva uses sfHeight (includes any portion of the foundation wall above grade). Do we want to combine these into one input to minimize ambiguity?

chipbarnaby:
> @chipbarnaby I have some input related questions for you:
> 
> * TOP soilDiff: We now need conductivity, density, and specific heat for Kiva. Do we still want to have this input if we can calculate it?

There is potentially a backwards compatibility issue.

Perhaps should leave soilDiff and make its default be the computed value.  If input, then verify that its value is consistent with computed value iff any of the other inputs is present.  Could get too complicated.

Are the conductivity, density, and specific heat values for Kiva also in TOP?

> * SURFACE sfDepthBG: Kiva uses sfHeight (includes any portion of the foundation wall above grade). Do we want to combine these into one input to minimize ambiguity?

The height above and below grade need to be known separately, right?  Then we need 2 inputs one way or another, I would think.  Unless I am missing something.

nealkruis:
I guess I'm wondering if we want to start also thinking about phasing out the old model that uses these inputs. We can support them for now, but issue a warning that perhaps they want to use the new ground model?

nealkruis:
@chipbarnaby do you want to look this over any more before we merge?

nealkruis:
@chipbarnaby Where does this branch stand? Any harm in merging it in?

chipbarnaby:
Probably OK, I should look at it.

PR #38, opened 2018-09-21
Added water heater wsTUse re-heat (whxBU) end use assignment

chipbarnaby:
Allow user specification of end use for accounting of virtual DHW resistance heater energy.  Default end use is dhwBU, so change is backward compatible.  With additional into, "real" backup (e.g. HPWH resistance heat) can be distinguished from energy used to maintain wsTUse.

Also some incomplete / harmless groundwork for volume vs duration draw modeling.


chipbarnaby:
Looks good!

PR #44, opened 2018-12-31
Add ability to model PV Power Electronics

nealkruis:


nealkruis:
@chipbarnaby if we remove the PVWatts DLL, we can do it on this branch.

src/CNRECS.DEF
```diff
@@ -3807,7 +3813,7 @@ RECORD PVARRAY "PVArray" *RAT		// input / runtime photovoltaics array
   *r FLOAT pv_convRatio		// ratio of back convection to front convection
   *r FLOAT pv_tGrndRatio	// ratio of ground-cell temperature diff. to air-cell temperature diff.
   *r *hide FLOAT_GEZ pv_thermCap  // thermal capacitance of the array, J/kg-m2
-  *h *e FLOAT pv_poaPv		// previous timestep plane of array incidence, Btu/h-ft2
+  *h *e FLOAT pv_radIPv		// previous timestep total radiation incident on array, Btu/h-ft2
```
chipbarnaby: The Pv suffix (for previous value I assume?) is unfortunate given that everything is pv here.  Elsewhere, we use Ls (last step) for this.

Also, is the timestep hour or subhour?

nealkruis: I agree. I think I borrowed the suffix convention from [here](https://github.com/cse-sim/cse/blob/2e08c7476442ff06578fbc19ff67d7194e659438/src/CNRECS.DEF#L491). I'm fine changing it. Currently PV is calculated hourly, but it could change to subhour down the road I suppose.

chipbarnaby: I looked at other contexts ... sometimes PvHr (for hourly values), sometime ls for subhour.  Not particularly consistent.  So perhaps avoiding Pv here would be helpful?  Not a big deal obviously.

src/SHADING.CPP
```diff
@@ -609,6 +609,9 @@ int SURFGEOM::gx_CalcBeamShading(		// beam shading for current hour
 #endif
 				fBeam = 1.;		// limit to 1
 			}
+			// round to nearest thousandth to minimize pixel precision 
+			// errors for unshaded surfaces and differences among GPUs
+			roundNearest(fBeam, 0.001f);
```
nealkruis: @chipbarnaby thoughts on this? I think it will eliminate hardware differences.

PR #50, opened 2019-01-23
Fix sky temperature calculation

### LINKED ISSUE #19: Investigate sky temperature algorithm
Standard 140 revealed noticeable differences in sky temperature between CSE and other programs on cloudy days.

I noticed a significant difference between the ACM documentation and the source code. 

In calculating sky emissivity, the ACM (equation 76) shows:

eSky = e0 + 0.784*(1 - e0)*cldCvr/10

While the [source code](https://github.com/cse-sim/cse/blob/cbf478d36cd776a42c6d47d291eb0cc7db34f0f6/src/WFPAK.CPP#L961-L972) shows:

eSky = e0 + 0.467*(1 - e0)*cldCvr/10

As a point of comparison, TRNSYS uses:

eSky = e0 + 0.9*(1 - e0)*cldCvr/10

The "test program" reference results in Standard 140 use the same methodology as described in the ACM for equation 73 (which, to my knowledge is not actually used in CSE):

screen shot 2017-11-09 at 11 36 03 am

nealkruis:
Fixes #19.


chipbarnaby:
I have confirmed that the test results do not change.  They all use CSW
weather files.

We should change some tests to use epw or add new epw-specific tests.

PR #58, opened 2019-04-09
Add DOE2 and TARP convection models for 140 comparisons

nealkruis:


src/CNCULT3.CPP
```diff
@@ -2170,23 +2171,29 @@ void SBC::sb_SetRunConstants(		// set mbrs that do not change during simulation
...
-			hcNatConstHtFlowUp = hcNatASHRAE[ xsty][ 1];
+			hcNatConstHtFlowReduced = hcNatASHRAE[ xsty][ 0];
+			hcNatConstHtFlowEnhanced = hcNatASHRAE[ xsty][ 1];
 			sb_hcConst[ 2] = Top.tp_hConvF * 0.88;
 		}
-		else
-		{	// values for C_CONVMODELCH_UNIFIED
+		else if (sb_hcModel == C_CONVMODELCH_UNIFIED)
+		{	
 			// set up but not used for other
```
nealkruis: @chipbarnaby results are changing in the Attic test cases because the debug reports print the sb_hcConst in the else statement here (note: "set up but not used for other"). The INPUT hc model now falls under the new else statement which does something different, but still unused in the INPUT case.

The other diffs are related to scientific notation in some of the results. Not sure why it's happening now and only for two test cases.

nealkruis:
@chipbarnaby regarding std::abs(), see https://stackoverflow.com/a/3118188/1344457. Maybe we should be using this throughout?

PR #60, opened 2019-04-11
Update toolset to v141_xp

nealkruis:


nealkruis:
@chipbarnaby fixed conflicts. Also added basic CI support.

PR #69, opened 2019-07-30
Use Git submodules instead of CMake external projects for dependencies

nealkruis:
This should 🤞 make handling dependency repositories easier.


nealkruis:
@chipbarnaby This is ready to review.

PR #96, opened 2020-05-09
Fix single axis tracker algorithm

nealkruis:


src/PVCalc.cpp
```diff
@@ -303,35 +303,44 @@ RC PVARRAY::pv_CalcPOA()
...
-		float sinz = sin(acos(cosz));
-		float x = (sinz*sin(azm - pv_azm)) / (sinz*cos(azm - pv_azm)*sin(pv_tilt) + cosz*cos(pv_tilt));
-		float psi;
-		if (x < 0.f && (azm - pv_azm) > 0.f) {
+		// Based on Rotation Angle for the Optimum Tracking of One-Axis Trackers by William F.Marion and Aron P.Dobos
+		const float sinz = sin(acos(cosz));  // sin of zenith angle
+		float azm_delta = azm - pv_azm;
+
+		// normalize azm_delta between -Pi and Pi
+		if (azm_delta >= kPi)
```
chipbarnaby, line 311: Normalization test s/b > not >=.  if azm_delta == kPi, the code would subtract k2Pi, yielding -kPi, then add k2Pi, yielding kPi.  Not wrong, but ... .  Should also use else.
if ( d > kPi)
   d-= k2Pi
else if (d < -kPi)
  d+=k2Pi

PR #144, opened 2021-12-15
Remove XMODULE file and DLL capabilities

galanCA:
This pull request comments out XMODULE from the build and adds a macro to remove XMODULE dependencies. By removing XMODULE, also removes share library interface features from CSE.


src/CGSOLAR.CPP
```diff
@@ -484,6 +484,7 @@ void ZNR::zn_SGClearTots()		// clear SG totals
 	}
 }		// ZNR::zn_SGClearTots
 //-----------------------------------------------------------------------------
+
```
nealkruis, line 487: Not a big deal here, but whitespace changes should be intentional. I suspect these are incidental from reverting earlier changes. There are a lot of them in this PR and can distract from the actual changes that need reviewing.

src/CNGLOB.H
```diff
@@ -84,6 +84,9 @@ typedef long long LLI;		// signed 64 bit integer
 typedef unsigned long long ULLI;
 //typedef int BOOL;	if needed: // 16 or 32 bit Boolean, matches windows.h.
 
+/*---------------------------- Remove DLL Support ---------------------------*/
+#undef SUPPORT_DLLS
```
nealkruis, line 88: I think we can do something a little more descriptive and specific than `SUPPORT_DLLS`. Maybe `ELIMINATE_XMODULE`?

nealkruis (changes requested):
Looks good! I left some minor comments. The only one that I think would hold this up is the naming of the pre-processor variable. @chipbarnaby, any thoughts?

src/CGSOLAR.CPP
galanCA: Ok, I'll go back and delete the white spaces.

cmake/configure.cmake
```diff
@@ -2,7 +2,7 @@ message("Making build directory: ${CMAKE_CURRENT_SOURCE_DIR}/msvc/build")
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/msvc)
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/msvc/build)
 message("Generating project files...")
-execute_process(COMMAND ${CMAKE_COMMAND} ${CMAKE_CURRENT_SOURCE_DIR} -T v142 -A Win32
+execute_process(COMMAND ${CMAKE_COMMAND} ${CMAKE_CURRENT_SOURCE_DIR} -T v142 -A Win32 -D CMAKE_SYSTEM_VERSION=10.0.17763.0
```
nealkruis, line 5: This is new! I don't think it's relevant to this PR.

src/CNCULT3.CPP
```diff
@@ -2420,7 +2420,6 @@ void XSURF::xs_SetRunConstants()
 	{	xs_sbcI.sb_SetRunConstants(dbPrint);
 		xs_sbcO.sb_SetRunConstants(dbPrint);
 	}
-
```
nealkruis, line 2423: Missed whitespace change?

src/ashwface.cpp
```diff
@@ -1062,8 +1062,6 @@ const float absS[] = { 0.f, 10.f, 20.f, -1.f };
 }
 #endif	// _DEBUG
 //=============================================================================
-
```
nealkruis, line 1065: Missed whitespace change.

src/ashwface.cpp
```diff
@@ -1379,7 +1383,4 @@ const CFSTYX* XASHWAT::xw_FindLibCFSTYX(
 	return pClosest;
 }		// XASHWAT::xw_FindLibCFSTYX
 //=============================================================================
-
```
nealkruis, line 1382: Missed whitespace change.

src/ashwface.h
```diff
@@ -49,7 +49,6 @@ struct CFSTYX : public CFSTY
 	float SHGCcogNFRC;	// NRFC cog SHGC (ditto)
 	float UcogAW;		// ASHWAT cog U
 	float SHGCcogAW;	// ASHWAT cog SHGC
-
```
nealkruis, line 52: Missed whitespace change.

nealkruis (changes requested):
See comments below.

src/CNGLOB.H
nealkruis: Let's go with `SUPPORT_XMODULE` so we don't have to change all the references to be negative. @chipbarnaby let us know if you have any other concerns with this PR otherwise we'll merge.

src/XMODULE.h
```diff
@@ -5,7 +5,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // XMODULE.h -- interfaces to external DLL/EXE functions
 ///////////////////////////////////////////////////////////////////////////////
-
+#if defined(SUPPORT_DLLS)
```
chipbarnaby, line 8: Should this be SUPPORT_XMODULE?

src/ashwface.h
```diff
@@ -292,8 +292,11 @@ class FENAW
 #define FCSL( s) s, sizeof( s)
 #define FCSET( d, s) strSpacePad( FCSL( d), s)
 #define FCGET( s) strTrim( NULL, FCSL( s))
+#if defined(SUPPORT_XMODULE)
 #include "xmodule.h"
 class XASHWAT : public XMODULE
+#endif // SUPPORT_XMODULE
+class XASHWAT
```
chipbarnaby, line 299: This needs an #else.  Would not compile if SUPPORT_XMODULE is defined.

In fact, did you confirm that the code will build with #define SUPPORT_XMODULE?

chipbarnaby (changes requested):
I added a couple of comments -- minor changes still required.

src/ashwface.h
galanCA: Thank you for catching that error. I build the code with #define SUPPORT_XMODULE and the only thing that one needs to add is the xmodule.cpp and xmodule.h on the CMakeList. Once that is added, it should run.

src/CMakeLists.txt
```diff
@@ -196,10 +198,17 @@ set(headers
...
-  XMODULE.h
   YACAM.H
 )
 
+if (USE_XMODULE)
+  list(APPEND source XMODULE.CPP) # Runtime Share Library interface - Not used since Version 0.899. (Windows only)
+  list(APPEND headers XMODULE.H) # Runtime Share Library interface - Not used since Version 0.899. (Windows only)
+  message(STATUS "Adding xmodule")
+else()
+  message(STATUS "Ignode xmodule")
```
nealkruis, line 209: These messages aren't necessary.

src/CNGLOB.H
```diff
@@ -84,6 +84,15 @@ typedef long long LLI;		// signed 64 bit integer
...
 //typedef int BOOL;	if needed: // 16 or 32 bit Boolean, matches windows.h.
 
+/*---------------------------- Remove DLL Support ---------------------------*/
+#if USE_XMODULE
+#define SUPPORT_XMODULE
+#pragma message("CNGLOG: SUPPORT_XMODULE is defined")
+#else
+#undef SUPPORT_XMODULE
+#pragma message("CNGLOG: SUPPORT_XMODULE is undefined")
+#endif // USE_XMODULE
```
nealkruis, line 94: Can you think of a way to handle this where you don't need to define use both `USE_XMODULE` and `SUPPORT_XMODULE`?

Also, we don't need the `#pragma` messages. If the user chooses to build with XMODULE support it will be a fairly conscious decision and they don't need to be reminded of it.

src/XMODULE.cpp
```diff
@@ -7,6 +7,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 #include "cnglob.h"
+#if defined(SUPPORT_XMODULE)
```
nealkruis, line 10: I don't think these are necessary either. If XMODULE isn't part of the CMake executable, these won't do anything.

src/XMODULE.h
```diff
@@ -5,7 +5,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // XMODULE.h -- interfaces to external DLL/EXE functions
 ///////////////////////////////////////////////////////////////////////////////
-
+#if defined(SUPPORT_XMODULE)
```
nealkruis, line 8: Same here.

nealkruis (changes requested):
While these changes work, there are a couple of changes to handle this more cleanly. Let's connect via Slack and Meet to go over them.

One thing that is concerning here is changes I'm seeing in HPWHsim and Kiva submodules. Those shouldn't be changing on this branch. Submodules are tricky and I'm happy to go over these with you.

src/CMakeLists.txt
```diff
@@ -196,10 +198,14 @@ set(headers
   VRPAK.H
   WFPAK.H
   XIOPAK.H
-  XMODULE.h
   YACAM.H
 )
 
+if(USE_XMODULE)
+  list(APPEND source XMODULE.CPP) # Runtime Share Library interface - Not used since Version 0.899. (Windows only)
+  list(APPEND headers XMODULE.H) # Runtime Share Library interface - Not used since Version 0.899. (Windows only)
```
nealkruis, line 206: Probably don't need to repeat the comment twice any more since they are right next to each other now. Maybe move it above on its own line?

nealkruis (approved):
Looks good to me!

chipbarnaby (approved):
Looks fine to me.
Bringing this option all the way out to cmake is perhaps overkill, but it is good to have a clean example of how to do that.

PR #145, opened 2021-12-15
Runs on Ubuntu

galanCA:
By eliminating windows-only headers from the code, CSE is no longer windows specific. This is a step in creating a multi-platform code.


galanCA:
Create a new pull request which will replace this one. The new pull request branches off from "Check for free space in disk drive free space function.". While this pull request will be designed to be submitted once CSE runs successfully on both Windows and Linux.

galanCA:
The team has opted to create more minor pull requests to move CSE into a multiplatform program instead of doing it all at once. This pull request is not needed anymore.

PR #146, opened 2021-12-16
Eliminate Windows-only headers

galanCA:
## Description

By eliminating windows-only headers from the code, CSE is no longer windows specific. This is a step in creating a multi-platform code. It uses a macro called eliminate windows to remove all windows.h and direct.h headers from the code. In terms of CSE behavior, nothing should change, and it should work just like it did before. For a test, I build and test CSE with `ELIMINATE_WINDOWS` on and off. 

## Author Progress Checklist:

- [x] Open draft pull request
    - [x] Add at least one label
- [x] Self-review of code
    - [x] I have performed a self-review of my code
    - [x] I have made corresponding changes to the documentation
        - [x] exercising the code changes in the test framework, and/or

## Reviewer Checklist:

 - [ ] If you are the last reviewer to approve, merge the pull request and delete the branch


tanaya-bigladder (approved):
Heroic! 
Builds fine on my PC with a recent version of Windows SDK (19041.0).

src/CNGLOB.H
```diff
@@ -84,17 +84,28 @@ typedef long long LLI;		// signed 64 bit integer
 typedef unsigned long long ULLI;
 //typedef int BOOL;	if needed: // 16 or 32 bit Boolean, matches windows.h.
 
-#undef ELIMINATE_WINDOWS		// #define to eliminate Windows dependencies
-								//   work in progress, 10-2017
-#if defined( ELIMINATE_WINDOWS)
-// typedefs based on windows.h
+/*---------------------------- Remove DLL Support ---------------------------*/
+#undef SUPPORT_DLLS
```
nealkruis, line 88: This will always undefine `SUPPORT_DLLS` even if a user defines it in CMake.

src/CNGLOB.H
```diff
@@ -84,17 +84,28 @@ typedef long long LLI;		// signed 64 bit integer
...
+
+#if 1
+#define	ELIMINATE_WINDOWS
+#else
+#undef	ELIMINATE_WINDOWS		// #define to eliminate Windows dependencies
+#endif 								//   work in progress, 10-2017
```
nealkruis, line 94: You don't need the if-statement here. Instead, you can have one line with `#define ELIMINATE_WINDOWS` that you can comment out if you don't want it defined.

src/ENVPAK.CPP
```diff
@@ -91,13 +102,15 @@ t }
 //=============================================================================
 WStr enExePath()		// full path to current executable
 {
-	char exePath[_MAX_PATH + 1];
 	WStr t;
+	#if !defined( ELIMINATE_WINDOWS)
```
nealkruis, line 106: The current convention is to no indent preprocessor statements.

src/ENVPAK.CPP
```diff
@@ -91,13 +102,15 @@ t }
...
+	char exePath[_MAX_PATH + 1];
 	if (GetModuleFileName(NULL, exePath, sizeof(exePath)) > 0)
```
nealkruis, line 108: We should be able to write a function that gets the executable path on any platform. I believe I shared what EnergyPlus is doing for this.

src/RCDEF.CPP
```diff
@@ -609,27 +622,31 @@ int CDEC main( SI argc, char * argv[] )
...
-		if (chdir(incdir))
+		filesys::path incdirTemp = filesys::path(incdir);
+		filesys::current_path(incdirTemp, pathEC);
+		if (pathEC)
 		{
 			printf( "\nCannot find include directory '%s'.\n", incdir);
 			Errcount++;
 		}
-		chdir(dbuff);
+		filesys::current_path(dbuff);
```
nealkruis, line 634: Let's go over these changes on a call. I think we can come up with a cleaner solution.

src/RCDEF.CPP
```diff
@@ -609,27 +622,31 @@ int CDEC main( SI argc, char * argv[] )
...
-		if (chdir(cFilesDir))
+		filesys::path cFilesDirTemp = filesys::path(cFilesDir);
+		filesys::current_path(cFilesDirTemp, pathEC);
+		if (pathEC)
 		{
 			printf( "\nCannot find include directory '%s'.\n", cFilesDir);
 			Errcount++;
 		}
-		chdir(dbuff);
+		filesys::current_path(dbuff);
```
nealkruis, line 649: Same here...though I wonder if we want to just make a function since this is nearly identical to the last set of changes.

src/RMKERR.CPP
```diff
@@ -563,9 +566,13 @@ RC errI(		// report error common inner fcn for errCrit, errV
 	const char* text;
 	if (erOp & SYSMSG)
 	{
+		#if !defined( ELIMINATE_WINDOWS)
```
nealkruis, line 569: I'd like to understand a bit more of the context here before we just cut out this code.

src/STRPAK.CPP
```diff
@@ -1125,10 +1125,10 @@ char* strPluralize(				// form plural of a word
 		{
 			strcpy( d, word);
 			int len = strlen( d);		// len >= 1 (check above)
-			BOOL bLastUpper = IsCharUpper( d[ len-1]) > 0;
+			BOOL bLastUpper = isupperW( d[ len - 1] ) > 0;
 			// standard rules y -> ies, else add s
-			if (tolower( d[ len-1]) == 'y')
-				strcpy( d+len-1, bLastUpper ? "IES" : "ies");
+			if (tolower( d[ len -1]) == 'y')
+				strcpy( d+len -1, bLastUpper ? "IES" : "ies");
```
nealkruis, line 1131: Whitespace changes?

src/STRPAK.CPP
```diff
@@ -1181,10 +1182,10 @@ char* strCase( char* s, const char toCases[3])
 		else				// spaces don't need case conversion
 		{
 			if (itc[which]==upper)
-				CharUpperBuff( s+i, 1);
+				*((unsigned char*)s+i) = toupper(*((unsigned char*)s + i));
```
nealkruis, line 1185: I want to make sure we understand what this code is doing. I'm not sure I do yet.

src/STRPAK.H
```diff
@@ -28,12 +28,7 @@ const int STRPPEXT   = 8;
 inline int fcn##W( int c) { return fcn( c); } \
 inline int fcn##W( char c) { return fcn( (unsigned char)c); } \
 inline int fcn##W( unsigned char c) { return fcn( (unsigned char)c); }
-// cause error in VS2008 (won't compile in VS6 due to Dundas grid STL troubles, 5-31-06)
-#if _MSC_VER >= 1400
 #define charTypeErr( fcn) Should_use_##fcn##W
```
nealkruis, line 31: I think we should add a comment here to help avoid confusion in the future. Something like, "`fcn##W is needed to handle type conversions that the original functions get tripped up on."

src/TIMER.CPP
```diff
@@ -36,7 +36,9 @@ static double tmrTicksToSecs(
 #if defined( TIMING_HIRES)
 
 // hi-res timing (Windows dependent)
+#if !defined( ELIMINATE_WINDOWS)
 #include <windows.h>
+#endif
```
nealkruis, line 41: Can we just remove this include altogether?

src/XIOPAK.CPP
```diff
@@ -176,9 +197,11 @@ RC FC xfGetPath(	// Check path / retrieve full pathname of a file
 
 // Returns RCOK if ok, with *fnamp replaced, else RCBAD.
 {
-	const char *fname, *p, *namepart, *dir;
```
nealkruis, line 179: These are similar to other changes in RCDEF. Let's go over them on a call.

src/XIOPAK.CPP
```diff
@@ -650,19 +664,17 @@ LI FC xdisksp()		// returns diskspace on default drive
...
 
-	if ( !GetDiskFreeSpace( rootPath, &spc, &bps, &freec, &totcl))
+	if ( !spaceErrorCode )
 		// GetLastError for more info
 		return 0L;				// call failed, return 0 space avail
 
-	if ( (float)spc * (float)bps * (float)freec < (float)LONG_MAX)	// if free space fits in a long
-		return spc * bps * freec;					// compute fixed point free space & return it
+	if ( freeSpace.available < (float)LONG_MAX)	// if free space fits in a long
+		return freeSpace.available;					// compute fixed point free space & return it
```
nealkruis, line 677: We talked about testing this, but I didn't see any mention of that in your PR description.

src/BRFW.CPP
```diff
@@ -9,8 +9,10 @@
 #ifdef WINorDLL
 #if 0
 #define STRICT
+#if !defined( ELIMINATE_WINDOWS)
 #include <windows.h>
 #endif
+#endif
```
nealkruis, line 15: Can we remove this include altogether?

src/CNGUTS.CPP
```diff
@@ -30,7 +30,9 @@
 #include "cse.h"
 #include <impf.h>	// impfStart impfIvl 2-94
 #ifdef WINorDLL
+#if !defined( ELIMINATE_WINDOWS)
 #include <windows.h>		// probably needed before brfw.h
+#endif
```
nealkruis, line 35: Can we remove this include altogether?

src/RMKERR.CPP
```diff
@@ -42,6 +42,9 @@
 
 #include <rmkerr.h>		// decls for this file
 
+#if defined( ELIMINATE_WINDOWS)
+#define NO_ERROR 0
```
nealkruis, line 46: Where was this defined before?

nealkruis (changes requested):
This is good progress, @galanCA! Let's find some time to go over some of the details that I comment on.

src/ENVPAK.CPP
galanCA: In order to use the EnergyPlus implementation, I need to include `windows.h`. Once I do that, I run into the same problem where some of `windows.h` definitions are included in CNGLOB.H.

Specifically lines [100](https://github.com/cse-sim/cse/blob/remove-windows-headers/src/CNGLOB.H#L100), [106](https://github.com/cse-sim/cse/blob/remove-windows-headers/src/CNGLOB.H#L106), and [107](https://github.com/cse-sim/cse/blob/remove-windows-headers/src/CNGLOB.H#L107).

Should I go ahead and change the definitions to CSE_BOOL and so on?

src/XIOPAK.CPP
galanCA: How can I add the test file where I prove they both have equal behavior for reference?

src/ENVPAK.CPP
nealkruis: Let's do this at the very end. I'm thinking that if we do a change, we might want to just use the standard data types (e.g., `bool`. 

@chipbarnaby Any thoughts?

src/XIOPAK.CPP
nealkruis: Good question. CSE isn't yet set up with unit tests (that would be the ideal approach here). I think it's fine just to mention your tests in the PR description. Our conversation here can also help document that you tested these changes and that I reviewed them with you.

src/ENVPAK.CPP
```diff
@@ -7,9 +7,16 @@
 /*------------------------------- INCLUDES --------------------------------*/
 #include "cnglob.h"
 // #include "cse.h"
+#if !defined( ELIMINATE_WINDOWS)
+#include <windows.h>
+#include <direct.h>	// getcwd
```
nealkruis, line 12: These will eventually need to be wrapped in an `#if (_WIN32)`.

src/ENVPAK.CPP
```diff
@@ -7,9 +7,16 @@
...
+#endif
+
+#ifdef linux
```
nealkruis, line 15: `linux` is deprecated. Use `__linux__` instead.

src/RMKERR.CPP
nealkruis: @chipbarnaby, after looking at this a bit closer I'm not sure we'll ever hit an instance with a SYSMSG. Maybe we can remove this code entirely?

src/STRPAK.H
nealkruis: The idea here is to explain what's happening on line 33. I like the direction you are going, but I think it might be more accurate to say:

> `charTypeErr` is used to redefine the original functions so that when they are used, the compiler will throw an error because "Should_use_##fcn##W" is not defined. This indicates to the user that they should use the "W" suffixed version of the function instead for proper casting.

src/XIOPAK.CPP
```diff
@@ -34,7 +34,9 @@
 /*------------------------------- INCLUDES --------------------------------*/
 
 #include <io.h>		// open, read, write,
+#if !defined( ELIMINATE_WINDOWS)
 #include <direct.h>	// getcwd, chdir, _chdrive
+#endif
```
nealkruis, line 39: Can we remove this entirely? With the move to file_system, I don't think anything in this file uses direct.h anymore.

nealkruis (changes requested):
This is very close now! Just a few more comments to address.

src/XIOPAK.CPP
galanCA: Yes, we can. I'll remove it.

src/ENVPAK.CPP
galanCA: Ok, good to know for the future.

src/ENVPAK.CPP
galanCA: The `direct.h` can go but yes. Eventually, we can have 
```
#if (_WIN32)
#include <windows.h>
#else 
#include <unistd.h>
#endif
```
Since they are both used for `enExePath`.

src/XIOPAK.CPP
```diff
@@ -176,64 +176,32 @@ RC FC xfGetPath(	// Check path / retrieve full pathname of a file
 
 // Returns RCOK if ok, with *fnamp replaced, else RCBAD.
 {
-	const char *fname, *p, *namepart, *dir;
-	SEC sec;
-	char save, cwdSave[ _MAX_PATH];			// save current dir pathname
-
+	filesys::path fname(*fnamp), dir;	// filesystem variables
```
nealkruis, line 179: Is `dir` an appropriate name still?

src/XIOPAK.CPP
```diff
@@ -176,64 +176,32 @@ RC FC xfGetPath(	// Check path / retrieve full pathname of a file
...
-			return RCBAD;			// give bad return
+	if (filesys::exists(fname, ec)) {	// File or full path exist
+		dir = filesys::absolute(fname); // Create full path
 	}
-	else	// no drive, no directory
-	{
-		namepart = fname;
-		dir = xgetdir();		// get full current drive/dir
+	else {
+		if (!fname.has_root_name()) {			// Missing root name, ex C:
```
nealkruis, line 186: I think we should understand the values of `ec` that are returned and use it directly instead. We maybe don't even need each of these separate checks because the `std::error_code` should contain information that we can pass back to the user. This is probably the case for many of our uses of `std::error_code`.

nealkruis (changes requested):
I added a few comments here. Let me know if you want to have a call to discuss them.

src/XIOPAK.CPP
```diff
@@ -176,64 +174,22 @@ RC FC xfGetPath(	// Check path / retrieve full pathname of a file
...
-			return RCBAD;			// give bad return
+	if (filesys::exists(fname, ec)) {		// File or full path exist
+		absolutepath = filesys::absolute(fname); // Create full path
 	}
-	else	// no drive, no directory
-	{
-		namepart = fname;
-		dir = xgetdir();		// get full current drive/dir
+	else {
+		err(erOp, ec.message().c_str(), fname);
```
nealkruis, line 184: How did you test this? What happens when the path doesn't exist? What message do you get here?

src/XIOPAK.CPP
```diff
@@ -176,64 +174,22 @@ RC FC xfGetPath(	// Check path / retrieve full pathname of a file
...
+		return RCBAD;
 	}
 
-// assemble, with \ between dir and filename one is lacking
-	*fnamp = strtPathCat( dir, namepart);
+	// assemble, with \ between dir and filename one is lacking
+	*fnamp = strtPathCat(absolutepath.parent_path().string().c_str(), absolutepath.filename().string().c_str());
```
nealkruis, line 189: As discussed, I'm not sure the concatenation is necessary.

nealkruis (changes requested):

src/XIOPAK.CPP
```diff
@@ -643,28 +601,25 @@ SEC FC xfflush(		// Flush all buffers associated with a file
...
-	DWORD totcl;
+	// Determine the available space on the filesystem on which current path is located
+	std::error_code spaceErrorCode;
+	uintmax_t freeSpace = filesys::space(filesys::current_path(), spaceErrorCode).available;
 
-	if ( !GetDiskFreeSpace( rootPath, &spc, &bps, &freec, &totcl))
+	if ( !spaceErrorCode )
 		// GetLastError for more info
-		return 0L;				// call failed, return 0 space avail
+		return 0ui64;					// call failed, return 0 space avail
```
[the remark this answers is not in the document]
nealkruis: We agreed to just drop the suffix altogether.

src/XIOPAK.CPP
nealkruis: @galanCA did you look at other uses of `std::error_code` to see if we want to change the messages there too?

src/XIOPAK.CPP
galanCA: Here I am having trouble doing `*fnamp = absolutepath.string().c_str()` because when it does the filesystem::string it store it on a temporary variable that is destroyed as soon as `filesystem.c_str()` is use, which means that garbage is passed to `c_str()`. The solution is to have a temporary variable and do as shown below

`std::string temp = absolutepath.string();
*fnamp = temp.c_str();`

I believe this does not happen during the function because the value is stored in the input variable.  But I could not find anything to support this claim.

src/XIOPAK.CPP
galanCA: The message that I get is `The operation completed successfully.` I try to break it (get a different message) but I don't know how.

src/ENVPAK.CPP
```diff
@@ -111,6 +144,7 @@ WStr enExeInfo(		// retrieve build date/time, linker version, etc from exe
 	codeSize = 0;
 	WStr linkerVersion;
 
+	#if !defined( ELIMINATE_WINDOWS)
```
galanCA, line 147: @nealkruis What would you like me to do with this section of code. Leave it like it is?

src/CNGLOB.H
```diff
@@ -84,17 +84,26 @@ typedef long long LLI;		// signed 64 bit integer
 typedef unsigned long long ULLI;
 //typedef int BOOL;	if needed: // 16 or 32 bit Boolean, matches windows.h.
 
-#undef ELIMINATE_WINDOWS		// #define to eliminate Windows dependencies
-								//   work in progress, 10-2017
-#if defined( ELIMINATE_WINDOWS)
-// typedefs based on windows.h
+/*---------------------------- Eliminate Windows ---------------------------*/
+#define	ELIMINATE_WINDOWS			// #define to eliminate Windows dependencies
```
galanCA, line 88: Do we want to keep `ELIMINATE_WINDOWS` for this PR?

nealkruis: No. Let's get this PR to run without it (using `_WIN32` as necessary).

src/ENVPAK.CPP
nealkruis: `if _WIN32` for now.

galanCA: Ok. I'll get rid off all the ELIMINATE_WINDOWS macros and request you when I am done.

nealkruis (approved):
This looks good to me! @chipbarnaby It's in your court now.

src/BRFW.CPP
```diff
@@ -9,7 +9,6 @@
 #ifdef WINorDLL
 #if 0
 #define STRICT
-#include <windows.h>
```
chipbarnaby, line 12: I think this windows.h should be retained.  It is #if'd out.  Leaving it retains how the scheme is supposed to work.

src/CNGUTS.CPP
```diff
@@ -30,7 +30,6 @@
 #include "cse.h"
 #include <impf.h>	// impfStart impfIvl 2-94
 #ifdef WINorDLL
-#include <windows.h>		// probably needed before brfw.h
```
chipbarnaby, line 33: Possibly leave this one also.  Again, it is not being compiled until WINorDLL is revived.

src/ENVPAK.CPP
```diff
@@ -20,7 +24,15 @@
 
 #include <envpak.h>	// declares functions in this file
 
-
+#if __has_include(<filesystem>)
```
chipbarnaby, line 27: This filesystem passage seems to appear too many times.  However, I think the thing to do is merge, then look at encapsulating the filesystem interface.

src/BRFW.CPP
nealkruis: @chipbarnaby this was my suggestion, but I think you're right. There are uses of what looks like Windows functions inside the WINorDLL conditions. Let's add a separate issue to explore how/if to support WINorDLL in the future.

src/RCDEF.CPP
```diff
@@ -609,27 +617,31 @@ int CDEC main( SI argc, char * argv[] )
...
-		getcwd( dbuff, 80);
+		filesys::path dbuff = filesys::current_path();
 		cFilesDir = argv[10];
-		if (chdir(cFilesDir))
+		std::error_code ec;
+		filesys::current_path(filesys::path(cFilesDir), ec);
+		if (ec)
 		{
-			printf( "\nCannot find include directory '%s'.\n", cFilesDir);
+			printf("\n\"%s\": %s", cFilesDir, ec.message().c_str());
```
chipbarnaby, line 641: Did you test this type of error?  Does the message provide appropriate info?  The old message said "include directory", so the user had some clue about what the problem is.

src/TIMER.CPP
```diff
@@ -36,7 +36,6 @@ static double tmrTicksToSecs(
 #if defined( TIMING_HIRES)
```
chipbarnaby, line 36: Did you try building with TIMER_HIRES defined?

src/XIOPAK.CPP
```diff
@@ -46,7 +45,18 @@
 #include "tdpak.h"
 #include "cvpak.h"
 
+#if __has_include(<filesystem>)
```
chipbarnaby, line 48: Isn't this filesystem passage also in xiopak.h and thus not needed here?

src/RCDEF.CPP
nealkruis: I'm not sure that is the only possible error you can encounter from changing directories. If the path doesn't exist it will say something like: `"path/to/dir": Path not found`. Another possibility is that the directory doesn't have read access (maybe?). Maybe we can provide more context, like "ec.message() for cpp file directory argument"

src/XIOPAK.H
```diff
@@ -95,7 +105,7 @@ int findFile1( const char* drvDir, const char* fName, char* fPathChecked=NULL);
 BOO findFile( const char* fName, const char* path, char* fPathFound);	// searches . first. NULL path searches env path.
 
 const char * FC xgetdir( void);
-SEC    FC xchdir( const char *, int erOp=WRN);
+SEC    FC xchdir( filesys::path, int erOp=WRN);
```
chipbarnaby, line 108: This requires filesystem to be known externally.  Can't the old signature be retained (or a std::string variant) and suitably mapped to filesys::path within the functions?

Way back when, this code ran both on PCs and on a Data General minicomputer.  The idea of things like xiopak and envpak was to hide the OS difference -- exactly our problem here.

chipbarnaby (commented):
I added some specific comments.

One question (noted in the at least one comment) is whether you tested error conditions that trigger error handling code.  Some error messages have been changed.  Do they still report something helpful?

src/ENVPAK.CPP
galanCA: You are right. I reduce it to only appear in XIOPAK.H since all the other files call XIOPAK.H.

src/CNGUTS.CPP
galanCA: Done. I don't understand why is GitHub not updating it.

src/RCDEF.CPP
```diff
@@ -621,7 +621,7 @@ int CDEC main( SI argc, char * argv[] )
 		cFilesDir = argv[10];
 		if (!filesys::exists(filesys::path(cFilesDir)))
 		{
-			printf("\n'%s': Path not found.\n", incdir);
+			printf("\n'%s': C++ file output directory path not found.\n", incdir);
```
nealkruis, line 624: @galanCA there is an error on this line.

nealkruis (changes requested):

src/RCDEF.CPP
galanCA: Fix.

src/TIMER.CPP
nealkruis: @galanCA didn't we decide that we would add a comment next to the `#undef TIMING_HIRES` to indicate that it is no longer maintained?

nealkruis (changes requested):

chipbarnaby:
Better yet, add a #error within the #if defined( TIMING_HIRES)?

nealkruis (approved):
@chipbarnaby back to you!

chipbarnaby (commented):
filesystem include remains in xiopak.h.  I thought we discussed hiding it within xiopak.cpp?

Then resurrect getcwd() etc.

Also, have the various error cases been tested?  Are reasonable error messaged generated?

nealkruis:
@chipbarnaby

1. We discussed this, yes, but I thought we decided that filesystem is part of the standard library and platform neutral. There isn't really a need to hide it, right? (* I guess it's still "experimental" in some implementations, but we don't expect it to change). Maybe we talked about changing the call signature back to `char*` to see what pushed us to using `filesystem::path`?
2. Why would we resurrect getcwd when there is platform neutral version in filesystem?
3. Can you be more specific (e.g., call out line numbers)? @galanCA tested the output messages for most of the functions. In many cases, the contents of `std::error_code.message()` are more specific than the previous messages which assumed a reason for a non-zero return code.

chipbarnaby:
The reason to hide filesystem is to keep a big header localized.  Only a
few features are needed by CSE, so hide them behind public functions.

In the limit, exposing things like filesystem would lead to "include
<everything>" in every file.

Unless there is something really magic about filesystem::path, I'm not sure
the world at large needs to know about it.  We resurrect getcwd() to hide
filesystem::path.

Regarding error messages, my question is -- Did you see them happen?  Do
they read OK?  Or put another way, did the new code actually ever get
executed?  It would be unfortunate if rare error code paths have bugs and
that is not discovered for 2 years.

nealkruis:
Thanks, @chipbarnaby!

We'll give it a try.

src/CNGLOB.H
```diff
@@ -112,7 +113,6 @@ typedef HINSTANCE HMODULE;
 #define near
 #define FAR   far
 #define far
-#define pascal
```
nealkruis, line 111: These are all obsolete now. Let's remove them all (including `NEAR`, which GitHub isn't letting me select 🤪 ).

nealkruis:
@chipbarnaby back to you on this one.

src/ENVPAK.CPP
```diff
@@ -322,7 +345,7 @@ signal( SIGFPE, 			// floating point errors, and integer errors (changes 0:0) un
 	//    to ease reporting calling location (in which case aborts, no return). */
 
 //--- save current drive/dir to restore at exit
-	getcwd( cwdSave, _MAX_PATH);	// save current dir (including drive) in static for byebye to use at exit
+	cwdSave = xgetdir();	// save current dir (including drive) in static for byebye to use at exit
```
chipbarnaby, line 348: This will not work.  xgetdir returns a pointer into tmpstr which will (probably) not survive the entire run.

What is needed here is 1) leave cwdSave as it was (character array).  2) strcpy( cwdSave, xgetdir()) or words to that effect.

Or we change xgetdir() to take a pointer or return a std::string.

We should perhaps have a short call about tmpstr?

src/XIOPAK.CPP
```diff
@@ -122,42 +131,28 @@ const char * FC xgetdir()
 /* Returns pointer to current working drive/directory pathname in Tmpstr[] */
 
 {
-	char * p = strtemp( _MAX_PATH );	// alloc n+1 bytes tmp strng space, strpak.cpp
+	char * p = strtemp( FILENAME_MAX );	// alloc n+1 bytes tmp strng space, strpak.cpp
+	strcpy(p, filesys::current_path().string().c_str());
 
-	return getcwd( p, _MAX_PATH);		// seems to include drive.
+	return p;		// seems to include drive.
```
chipbarnaby, line 137: As noted elsewhere, perhaps have xgetdir return std::string?

src/ENVPAK.CPP
galanCA: I implemented the first option, returning cwdSave to a character array.

galanCA:
Should we delete this Path class as the file system replaces it? 

https://github.com/cse-sim/cse/pull/146/files#diff-e03750543f9834b7ad26c49b1915a92f4fe381bfb92aaf18fd560e0c0a32d9f9L69-L85

PR #163, opened 2022-02-14
Fix MSVC compiler version to one where all tests pass.

tanaya-bigladder:
Added "version=xx.xx" option to "-T" flag for CMake configuration. It sets the CMAKE_VS_PLATFORM_TOOLSET_VERSION cmake variable.


chipbarnaby (approved):
We'll find out if this works ...

PR #191, opened 2022-03-21
Fancoil

chipbarnaby:
Initial version of RSYS fancoil model


src/CNLOADS.CPP
```diff
@@ -2287,7 +2287,9 @@ RC RSYS::rs_TopRSys1()		// check RSYS, initial set up for run
 				rc |= oer("rsSEER (%g) must be > rsEER (%g)", rs_SEER, rs_EER95);
 
 			// checks to prevent crazy extrapolation results
-			rc |= rs_CkFRatio(RSYS_CAP82, RSYS_CAP95, RSYS_CAPRAT8295, 1.f, 2.f);
+			// cap82 > cap95 typically; occasional cap82 < cap95 examples seen so use 0.8
```
nealkruis, line 2290: typo? "so" -> "to"?

nealkruis (approved):
Saw one typo in the comments. At this point this PR is really just the VCHP capacity checks and fixing merge related issues from previous PRs, right?

src/CNLOADS.CPP
chipbarnaby: No, I mean "so" as in "thus"

Issue #101, opened 2020-07-17
Version detection issues


chipbarnaby:
**Badly formed version tag fails ugly**
If the version tag is not formatted correctly, the build compiles most everything and then fails.
Should fail immediately and say why.

**Multiple tags not detected**
If the current tip has multiple tags, not clear which one is fetched.  If one is badly formatted, build may fail even though valid tag is present.

tanaya-bigladder:
Does "not formatted correctly" correspond to an error in grabbing the Git SHA/branch/tag?

nealkruis:
Yeah, this is something that I would like to generalize for Atheneum type projects too. There are a couple of tasks here:

1. Instead of giving up and assigning "?.?.?" as the version number, CMake should probably fail.
2. The regular expressions need to be able to capture other build meta data from a tag too ([per semantic versioning](https://semver.org/#spec-item-10)). I think this will fail the pattern recognition.

tanaya-bigladder:
To rephrase my question, if I want to reproduce the build failure, what do I need to make fail? Does "not formatted correctly" usually mean that the `git rev-parse` failed?

tanaya-bigladder:
Opened PR #202

PR #202, opened 2022-03-29
Fix versioning

tanaya-bigladder:
Address Issue #101 .
Any failure to capture branch/tag/build information from git will now result in a failure when CSEVersion.cmake is executed (early in the build process). If the version doesn't conform to "`v<major>.<minor>.<patch>-<optional_prerelease_label>`" it will also fail in the CMake script.
The final version, with metadata, conforms to semver.org guidelines.


nealkruis:
@tanaya-bigladder what do you think about adding some build meta data to indicate a "dirty" build (i.e., with uncommitted changes)?

nealkruis:
@tanaya-bigladder what do you think about adding some build meta data to indicate a "dirty" build (i.e., with uncommitted changes)?

tanaya-bigladder:
Most builds are dirty, aren't they? Just the existence of a branch name other than "master" would indicate that generally. I don't think it's a lot of work, but is there value in it if we are generally careful about releases?

nealkruis:
@tanaya-bigladder, I would not say most builds are dirty. A branch off of master does not make the build dirty. It is only dirty if there are uncommitted changes in the working copy.

tanaya-bigladder:
I've never had the experience of needing this information, but I can add it here especially if @nealkruis thinks it'll be useful for atheneum-based libraries in the future.

tanaya-bigladder:
OK, a solution is checked in! It's not my most ideal (`git status --porcelain` has a guaranteed interface over time whereas `git diff --shortstat` might not) but it will work without adding another script to parse the output of status.

nealkruis (approved):

Issue #213, opened 2022-04-03
Intermittent regression test difference -- "-0" vs "-1"


chipbarnaby:
The attached file fails regression test compare intermittently.

What is known --
* Differences are only for "-0" vs "-1".
* Differences to date have only been seen for a single report at a single time (Dec. 3, hr 5, subr 1).
* With trap code active in cvpak.cpp (release build) (see notes there), the inconsistent behavior was not seen.
* When "coldec=0" is changed to "coldec=3" on the report column in question (qBal2), the problem not seen even with many attempts.

See cvpak.cpp, search for M0BUG.

Attached files --
wshp.cse   CSE file exhibiting issue
wshp_1.rep   Output from one run
wshp_2.rep   Output from another run (differs)
wshp_ok.cse   colDec=0 changed to colDec=3 on column qBal2
wshp_ok.rep   Output from  wshp_ok.cse

[wshpex.zip](https://github.com/cse-sim/cse/files/8405097/wshpex.zip)

tanaya-bigladder:
Some questions, @chipbarnaby:

I get the following errors when I run SHA 7b500e0b with your wshp.cse input file, could you help me resolve them?

```
Command line: .\wshp.cse
  Z:\Projects\CSE_issue_213\wshpex\wshp.cse run 001: Input
---------------
  rsType = WSHP
           ^
Z:\PROJECTS\CSE_ISSUE_213\WSHPEX\WSHP.CSE(2087): Error:
  S0020: V0039: 'WSHP' is not one of ACFurnace ACResistance ACPkgRoom
    ACPkgRoomFurnace ACPkgRoomResistance ASHP ASHPHydronic ASHPPkgRoom
    VCHP2 AC Furnace Resistance FanCoil
    (OR mispelled word or omitted quotes)
---------------
Z:\PROJECTS\CSE_ISSUE_213\WSHPEX\WSHP.CSE(2085): Error: RSYS 'rsys1':
    S0496: Can't give 'rsCapAuxH' when rsType is not ASHP, ASHPHydronic, ASHPPkgRoom, or VCHP2
---------------
Z:\PROJECTS\CSE_ISSUE_213\WSHPEX\WSHP.CSE(2085): Error: RSYS 'rsys1':
    S0496: Can't give 'rsCOP47' when rsType is not ASHP, ASHPHydronic, ASHPPkgRoom, or ASHPVC (VCHP2)
---------------
[...]

PR #214, opened 2022-04-04
Add explicit model for WSHPs

nealkruis:


nealkruis:
@chipbarnaby Can you take a quick look through my commit? I noticed the CI failed on the regression test for WSHP, but I can't reproduce this on my Windows machine. Perhaps it's related to #213?

chipbarnaby (approved):

PR #225, opened 2022-04-12
Build with minimum SDK version 19041.789 to incorporate printf rounding bugfix (requires 20348.0 in practice).

tanaya-bigladder:
The title says it. Should fix Issue #213 . (Explanation in Issue link.)
I had to pick a branch and a build platform where I could reliably reproduce rounding errors; not every combination seems to do it. I can sort of believe that a known bug in the SDK CRT could cause this.
After rebuilding with the newest SDK, I discovered 5 regression reference files that had incorrect rounding - in all of those cases, a floating point number (> 0.5, < 1.0) was being rounded down to 0. They have been updated as well.
(We can't use `-DCMAKE_SYSTEM_VERSION=10.0.19041.789` because CMake will allow a match to 10.0.19041.0)


chipbarnaby (approved):
Spectacular.

PR #233, opened 2022-04-26
Delete random generated garbage lines at the end of the file

galanCA:
This PR is to delete a random line created at the end of the file which seems to not be a problem in MSVC but the GCC compiler seems to not be able to ignore them.

This should help reduce the amount of review needed for the Multi-platform branch.


tanaya-bigladder (changes requested):
All the other source files contain CR-LF at the end, so these should be made consistent.

tanaya-bigladder (approved):

PR #235, opened 2022-04-28
Exclude git from documentation artifact.

galanCA:
Exclude the git folder from the documentation artifact.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,9 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: doc/output
+          publish_dir: |
+                        doc/output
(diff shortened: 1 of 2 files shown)


chipbarnaby (approved):
Pretty simple!

.github/workflows/release.yml
```diff
@@ -42,7 +42,9 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: doc/output
+          publish_dir: |
+                        doc/output
+                        !doc/output/.git
```
nealkruis, line 47: Are you sure you want to do this here?

galanCA: I was looking for a way to test that line but I didn't know-how. Is this not the same as build-and-test.yml?

nealkruis (approved):
Looks good. Thanks, @galanCA!

PR #256, opened 2022-05-06
Replace windows string compare function

galanCA:
## Description

Replaces windows dependent compare function, for a c-style standard function. The output will be the same as it was originally intended. Replaces strcmpi, and _stricmp for in-house-built strcmpi.

The author test that the inhouse function replaces the windows strcmpi by intentionally breaking the function and seeing the effects it creates.

Replaces:
- [strcmpi](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcmpi?view=msvc-170) and [stricmp](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stricmp-wcsicmp?view=msvc-170) for [_stricmp](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stricmp-wcsicmp-mbsicmp-stricmp-l-wcsicmp-l-mbsicmp-l?view=msvc-170)
- [strnicmp](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strnicmp-wcsnicmp?view=msvc-170) for [_strnicmp](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strnicmp-wcsnicmp-mbsnicmp-strnicmp-l-wcsnicmp-l-mbsnicmp-l?view=msvc-170)

## Author Progress Checklist:

- [x] Open draft pull request
- [x] Self-review of code

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve


src/STRPAK.CPP
```diff
@@ -1332,6 +1332,22 @@ BOOL strMatch(					// string match
 			bSeenDifSpace = TRUE;		// both space but different
 	}
 }			// strMatch
+//-----------------------------------------------------------------------------
+int strcmpi(		// Substitude windows strcmpi functions
```
tanaya-bigladder, line 1336: I believe that what Chip was suggesting was to make this function do two different things, but at compile-time. If we're on a Windows system, this function calls the existing `_stricmp`. If not, it uses the new function.

src/STRPAK.CPP
```diff
@@ -1332,6 +1332,22 @@ BOOL strMatch(					// string match
...
+	const char* char1,	// First string to be compare
+	const char* char2	// Second string to be compare
+)
+// Compares two string ignoring case sensitivity
+{
+	int sum{ 0 };
+	for (;; char1++, char2++) {
+		sum += tolower((unsigned char)*char1) - tolower((unsigned char)*char2);
+		if (sum != 0 || *char1 == '\0' || *char2 == '\0') {
```
tanaya-bigladder, line 1345: Good idea to short-circuit if sum != 0.

src/STRPAK.CPP
chipbarnaby: This is not quite what I was proposing.

Proposal #1 -- use the Microsoft name and implement our own if needed.  That is

#if !defined( _WIN32)
int stricmp(  whatever)
our own code
#endif

The idea is to avoid the forwarded call when on Windows (in all cases, in fact).

Proposal #2 -- map the Microsoft name to another implementation if needed

#if !defined( _WIN32)
#define _stricmp( s1, s2)  strcasecmp( s1, s2)   // use POSIX implementation
#endif

Or we change all the calls to strcasecmp() and do the reverse definition -- see https://stackoverflow.com/questions/3694723/error-c3861-strcasecmp-identifier-not-found-in-visual-studio-2008
#ifdef _MSC_VER 
//not #if defined(_WIN32) || defined(_WIN64) because we have strncasecmp in mingw
#define strncasecmp _strnicmp
#define strcasecmp _stricmp
#endif

The ideas are 1) don't make chained calls and 2) attempt to avoid implementing our own.

tanaya-bigladder: I'm of two minds about this. It's fine to try and eliminate chained calls, but I can't imagine the compiler isn't optimizing that out? I find the price you pay in navigation and confusion when you realize that a legitimate function name is not, in fact, calling the function you thought it was, is generally too high for my liking.

galanCA: For now, I will implement option 1 because it will be challenging to test option 2 since I would like to at least build the program in MAC or Linux.  I will leave a comment to implement option 2 if we decide to go that route.

tanaya-bigladder (approved):

nealkruis (changes requested):
I need a better description of the testing process to understand what was done. What does "intentionally breaking the function" mean? Does it mean you are actually changing the function, or are you just passing in arguments that will have a non-zero return? Maybe an example or two would help.

src/STRPAK.CPP
```diff
@@ -1332,6 +1332,40 @@ BOOL strMatch(					// string match
 			bSeenDifSpace = TRUE;		// both space but different
 	}
 }			// strMatch
+//-----------------------------------------------------------------------------
+#ifndef _WIN32
```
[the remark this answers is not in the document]
galanCA: Compiler issue because these functions are only in MSVC. I'll fix it.

src/STRPAK.CPP
```diff
@@ -1332,6 +1332,40 @@ BOOL strMatch(					// string match
 			bSeenDifSpace = TRUE;		// both space but different
 	}
 }			// strMatch
+//-----------------------------------------------------------------------------
+#if defined(__clang__) || defined(__GNUC__) && defined(__linux__)
```
nealkruis, line 1336: `#ifndef _MSC_VER`?

nealkruis (changes requested):

src/STRPAK.CPP
galanCA: I thought this was more explicit.

nealkruis: What if the compiler is not MSVC, clang, or GCC?

tanaya-bigladder: Because the `_str***` functions are specific to the Microsoft compiler, our redefine is relevant for (ALL != Microsoft), rather than for known specific compilers. I think I would use `#ifndef _MSC_VER` as well here. 
(But this is part of the complication of redefining a symbol with the same name as a proprietary one, which I personally _really_ don't like doing, but Chip's ok'ed.)

nealkruis (approved):

PR #257, opened 2022-05-06
Replace windows convert case functions

galanCA:
## Description

Updates windows functions that change the case of a string either upper (strupr) or lower (strlwr). As recommended by windows. It also creates a shadow function used in GCC and Clang for these functions.

### Test

A unit test is added to this PR to ensure their behavior stays the same.

## Author Progress Checklist:

- [ ] Open draft pull request
- [ ] Self-review of code

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve


chipbarnaby (changes requested):
I don't think this is the right approach, especially compareCaseStr().  We do not want to pay the price of constructing std::strings just to be able to compare them.  That often involves memory allocation.

I believe there are valid standard functions for all the null terminated string operations we want to do.  Actually, is it known that the current functions are not available via libraries on other platforms?  I believe we are using standard C functions that are generally widely supported.  In other words, why is this necessary at all (or so broadly necessary).

At some point we may convert to using std::string, at which point std::string operations can be used without unnecessary conversion.

galanCA:
I will implement a similar system as PR #256.

galanCA:
I will implement a similar system as PR #256.

test/unit/strpak.unit.cpp
```diff
@@ -28,3 +28,25 @@ TEST(strpak, compare_case_insensitive_string)
 	EXPECT_LT( _strnicmp(long_string_upper_case, long_string_wrong_case, 22), 0);
 	EXPECT_EQ( _strnicmp(long_string_upper_case, long_string_wrong_case, 10), 0);
 }
+
+TEST(strpak, convert_case_functions)
+{
+	// Set up strings
+	char uppercaseString[] = "this should be uppercase.";
+	char lowercaseString[] = "THIS SHOULD BE LOWERCASE.";
+	char randomCaseString[] = "Hello There ARE many case HERE!";
```
tanaya-bigladder, line 37: Perhaps the random case string could also put an upper-case letter in the middle of a word.

tanaya-bigladder, line 37: Also, "there are many case**s**" here. :)

test/unit/strpak.unit.cpp
```diff
@@ -28,3 +28,25 @@ TEST(strpak, compare_case_insensitive_string)
...
+	char numberString[] = "There are 12311 numbers (#).";
```
tanaya-bigladder, line 38: Should we also test number/letter mixed words, as you sometimes see in variable names, e.g. "ASHRAE205"

tanaya-bigladder (changes requested):

test/unit/strpak.unit.cpp
```diff
@@ -28,3 +28,25 @@ TEST(strpak, compare_case_insensitive_string)
...
+
+	// Test convert uppercase strings
+	EXPECT_STREQ( _strupr(upper_case_string), "THIS SHOULD BE UPPERCASE.");
+	EXPECT_STREQ( _strupr(mixed_case_string), "HELLO THERE ARE MANY CASES HERE!");
+	EXPECT_STREQ( _strupr(number_string), "TESTING NUMBERS: ASHRAE205, 1252345 (#).");
+
+	
+	// Test convert lowercase strings
+	strcpy(mixed_case_string, "Hello There ARE maNy cases HERE!");
+	strcpy(number_string, "Testing numbers: ASHRAE205, 1252345 (#).");
```
tanaya-bigladder, line 48: I would create a new char array here; the reason is that if someone wanted to update this function later on, they could mistakenly strcpy a longer string into an already allocated char array and have a memory leak.

tanaya-bigladder (dismissed):

chipbarnaby (changes requested):
There could be trouble here due to signed characters.  tolower / toupper take integers.  That means a char > 127 will become a negative value.  The standard work-around is the following or words to that effect.

char* _strlwr(char* s)
{
    char* tmp = s;

    for (;*tmp;++tmp) {
        *tmp = tolower((unsigned char) *tmp);
    }

    return s;
}


This character stuff has always been tricky, but I think defining _strupr and _strlwr is a good enough solution pending conversion to std::string.

I would flag these functions as ?MULTIPLATFORM? -- this will work but certainly not elegant.

chipbarnaby (approved):

nealkruis (approved):

nealkruis:
@tanaya-bigladder merging is currently blocked because of your "request changes". Can you re-review?

tanaya-bigladder (commented):
Approved; sorry it got out of sync.

PR #266, opened 2022-05-17
Update CNGLOB for multi-platform 

galanCA:
## Description

Updates CNGLOB.H so it can be compiled on all platforms. This should not create any changes in the windows version of CSE.

The changes made in this PR are:

- Moving the location of global constants
- Moving the location of function templates because it was used before declaration.

The following definitions are used by the nested headers in cnglob.h
```
/*------------------ Definitions used by CSE headers -----------------------*/
template< typename T, typename TX> inline T min(T a, TX b)
template< typename T, typename TX> inline T max(T a, TX b)
typedef void* DMP;	// Dynamic memory block pointer: ptr to any type, record struct, etc, of caller's
typedef SI RC;
const int EROP1 = 0x010000;	//  vrpak.h  dmpak.h  xiopak.h    [ratpak.h]
const int EROP2 = 0x020000;	//  vrpak.h  dmpak.h  [xiopak.h]  sytb.cpp
const int IGN = 0x0001;	// ignore: no message, silently continue execution
const int ABT = 0x0003;	// issue msg, get keypress, abort program.
```

The last four lines are inside a group called Error Actions and Options ("erOp" arguments) . I decided to keep everything together and move it all before the include statements. However, I think it would be good to create a new header with these definitions for aesthetics purposes.

## Author Progress Checklist:

- [x] Open draft pull request
- [ ] Self-review of code
        - [x] manually verifying the changes (as explained in the the pull request description above)
    - [x] Merge in main branch and address resulting conflicts and/or test failures.

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve


src/CNGLOB.H
```diff
@@ -178,6 +178,21 @@ typedef std::string WStr;
 						//   NOTE: < > required re rcdef build
 
 /*---------------------- global macros and defines -------------------------*/
+// useful constants
+const double kPi=3.14159265358979323846;
```
tanaya-bigladder, line 182: If these are entirely new, let's think about making them `constexpr`.

tanaya-bigladder, line 182: I'm curious why you had to relocate the constants, and then, while we're at it, does it make sense to make them constexpr?

galanCA: I had to relocate them because one of the variables was used before it was declared. Yes, I think it meets all the requirements to make them constexpr.

tanaya-bigladder: Could you post the error you got? I'm a little confused as to why it worked all this time and now doesn't.

galanCA: I will. To clarify it works on msvc, it just doesn't work on POSIX system.

galanCA: Screen Shot 2022-05-27 at 11 56 56 AM

tanaya-bigladder (approved):

src/RMKERR.H
```diff
@@ -172,6 +172,13 @@ template< typename T> void MDbPrintf( 		// debug print a matrix
 		VDbPrintf( oMsk, NULL, pRow, nCol, fmt);
 	}
 }			// MDbPrintf
+// debugging aid: warn if limits invoked
+template< typename T> inline T bracketWarn( T vMin, T v, T vMax)
```
tanaya-bigladder, line 176: What caused this guy to move to a different file?

galanCA: Yes, `CNGLOB` has [these includes](https://github.com/cse-sim/cse/blob/master/src/CNGLOB.H#L733-L738) almost at the end. The only thing is that in doing that, I will have to move some definitions around. An example is moving `typedef SI RC` before `#include rmkerr.h`.

galanCA: @nealkruis would it be ok to only move `#include rmkerr.h` while keeping the other set of `#include` at the bottom? Or should I move them all together?

Initially I was going to move them all together for organization, but it looks like it will take more effort than only moving `#include rmkerr.h` and I don't know if it is worth it.

nealkruis: Maybe a question for @chipbarnaby. I'm not sure why those "more nested includes" are so low in the file or if there is a need to preserve the order of them.

galanCA: They depend on some definitions earlier in the files. At first glance, it looks like it is possible to reorganize all of it.

nealkruis (approved):

chipbarnaby (approved):

galanCA:
The reason why they cannot be higher up is because the nested includes depend on these definitions.

`/*------------------ Definitions used by CSE headers -----------------------*/`
`typedef void* DMP;	// Dynamic memory block pointer: ptr to any type, record struct, etc, of caller's`
`typedef SI RC;`
`const int EROP1 = 0x010000;	//  vrpak.h  dmpak.h  xiopak.h    [ratpak.h]`
`const int EROP2 = 0x020000;	//  vrpak.h  dmpak.h  [xiopak.h]  sytb.cpp`
`const int IGN = 0x0001;	// ignore: no message, silently continue execution`
`const int ABT = 0x0003;	// issue msg, get keypress, abort program.`

I decided to keep everything together which is why we have a block of code before it. We could create a header file with all of these definitions and just have it before the "nested include".

We could remove the repeat includes, but I did check that the header guards are there (I even included one in `vecpak.h`) so we should not have problem repeating headers. 

I don't know about precompile headers, but I will look for them in CLANG and GCC.

PR #290, opened 2022-05-25
Ancrec upgrade

galanCA:
## Description

Upgrades ANCREC.cpp, the items that were changed are:
- Replace memicmp with _strnicmp. This is done based on the assumption that .cse file will not contain null characters.
- Adding guards to SRD.H

## Author Progress Checklist:

- [ ] Open draft pull request
- [ ] Self-review of code
        - [x] manually verifying the changes (as explained in the the pull request description above)
    - [x] Merge in main branch and address resulting conflicts and/or test failures.

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve


tanaya-bigladder (approved):

src/SRD.H
```diff
@@ -14,7 +14,8 @@
 
 
 //=============== DATA TYPES: dttab.cpp
-
+#ifndef SRD_H
```
nealkruis, line 17: Seems strange that the comment above is outside the guard when it clearly is related to the comments below it.

src/STRPAK.CPP
```diff
@@ -1332,6 +1332,22 @@ BOOL strMatch(					// string match
 			bSeenDifSpace = TRUE;		// both space but different
 	}
 }			// strMatch
+//-----------------------------------------------------------------------------
+#ifndef _WIN32
+int memicmp(			// Compares memory ingnoring case sensitivity
```
nealkruis, line 1337: Where is this used in the code? It might help to add a comment that explains and why this is necessary as opposed to the other string compare functions?

galanCA: `memicmp` is used in 7 different places in CSE. I am adding this function because the first encounter is [here](https://github.com/cse-sim/cse/blob/master/src/ANCREC.CPP#L1258). I added a comment detailing the difference between `stricmp` vs `memicmp`. There is a great explanation in [this post].(https://stackoverflow.com/questions/13095513/what-is-the-difference-between-memcmp-strcmp-and-strncmp-in-c)

nealkruis: I just want to confirm that we need `memicmp`, or if we can convert them all to `stricmp`. That is, are there places where the count is not also the length of the string?

nealkruis: @galanCA did you find an answer to this? Also, how is it different from `_strnicmp`? You added in #257?

nealkruis: That would make sense to me. @chipbarnaby any heartburn?

nealkruis (commented):
How did you confirm these changes were getting hit in the ctest runs?

Also, the description of the PR doesn't seem accurate any more.

nealkruis:
@galanCA so how are you confident the changes are working in the context of the files that are not covered in the tests?

galanCA:
I am confident `ancrec.cpp` will work as intended. To make sure `detfnc.cpp` will work, I wrote a unit test in which I test all the options "Jan, Feb, etc" and "January, February, etc" and the result is as expected.

nealkruis:
We should add datfcns to the unit test executable and explicitly add the unit test there.

galanCA:
@nealkruis Looking into the ways memicmp was used previously the only difference is that if a null character is added inside a .cse file memicmp will be able to read it. However modern editors seem to skip it and some don't let you add it if we are comfortable that no null character will be added to .cse file then this change is ok. Do you still think a unit test is necessary?

nealkruis:
Just so we're on the same page, `memicmp` is used in the following contexts:

- `getFileIX`: Retrieves a file index by searching through the list of established file names. It is used in two locations:
  - `svFname` when processing user input tokens in the form `line <n> "<filename>"`. @chipbarnaby is this used anywhere? Maybe I'm not fully understanding what this is doing.
  - `ppOpI` (pp.cpp) preprocessor `#include` file names.
- `findMember` (cuprobe.cpp) used to match record names when processing probe descriptions.
- `dStr2Mi` converting month string to month index
- `ppmId` (pp.cpp) searching for "$eof" to know when to stop preprocessing a file.

In each of these cases, the strings will either be intentionally null terminated at the appropriate length or a length value will cut the comparison short of any null characters. Note that some of the character strings passed into these functions may not be null terminated as indicated by the comments, but that shouldn't cause these changes to break.

@chipbarnaby does that sound right to you?

chipbarnaby (approved):

PR #320, opened 2022-06-15
Add itoa for linux system.

galanCA:
## Description

Replace the content in this section with:
- The motivation and context for this change (if it is not immediately clear from the title)
- If it fixes an open issue, specify the issue number (e.g., "fixes #XXXX")
- A summary of the behavior expected from this change
- A description of tests performed

## Author Progress Checklist:

- [ ] Open draft pull request
- [ ] Self-review of code

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve

### DIFF
--- a/src/VRPAK.H
+++ b/src/VRPAK.H
@@ -74,4 +74,30 @@ void FC vrInfoClean( VROUTINFO *voInfo);
 RC FC vrUnspool( VROUTINFO *voInfo);
 BOO FC isUsedVrFile( const char* fName);
 
+#ifdef __gnu_linux__
+char* itoa(int value, char* result, int base) {
+	// check that the base if valid
+	if (base < 2 || base > 36) { *result = '\0'; return result; }
+
+	char* ptr = result, *ptr1 = result, tmp_char;
+	int tmp_value;
+
+	do {
+		tmp_value = value;
+		value /= base;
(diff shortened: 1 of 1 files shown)


chipbarnaby (commented):
1. iota() is only used if vrName is NULL.  Which, based on grep, is never.  So better solution is to drop itoa() and return RCBAD if vrName is NULL.
2. Even if iota() is retained, we do not need the ability to handle any base other than 10.  So you could simply use some variant of sprintf().
3. Further, if a function such as iota() must be added, it should probably not be added locally in the file where it is used.  In this case, perhaps strpak?  We already have too many functions that are in the "wrong" place.  Although in this case, as I said, we don't need it at all.

My vote is just drop it.  Return RCBAD if vrName is NULL.  Don't need nmBuf either it appears.

chipbarnaby (changes requested):
As written, this code can change spl.sp_nVrh and then fail if !vrName.

It is a good idea to have no effect if the function fails -- either succeed or do nothing.  As you can see, in the original version, checks that could return RCBAD are done before anything is altered.

BUT, in this case, the easiest change is

   if (!vrName)
           vrName = "?";

vrName is not critical and further, there are no calls in CSE with vrName==NULL.  So just use "?" and carry on.

Of course, change the comment on the argument as well.

galanCA:
@chipbarnaby Why does `vrName` changes if it is a `const char`? Is that allow?

chipbarnaby:
vrName is a pointer to const char.  So you can change vrName but not what it points to.

If it compiles, you're OK

PR #357, opened 2022-12-01
Fix x64 build in other compilers

galanCA:
## Description

The changes performed were: 

## Author Progress Checklist:

- [ ] Open draft pull request
- [ ] Self-review of code

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve

### DIFF
--- a/src/RCDEF/rcdef.cpp
+++ b/src/RCDEF/rcdef.cpp
@@ -1151,7 +1151,7 @@ LOCAL void wChoices(            // write choices info to dtypes.h file.
 LOCAL void wDttab()     // write C++ source data types table dttab.cpp
 {
 // open working file dttab.cx
-	char buf[MAX_PATH];
+	char buf[CSE_MAX_PATH];
 	xfjoinpath(cFilesDir, "dttab.cx", buf);		// buf also used to close
 	FILE* f = fopen( buf, "w");
 	if (f==NULL)
@@ -1256,7 +1256,7 @@ LOCAL void wDttab()     // write C++ source data types table dttab.cpp
 // terminate file, close, update if different
 	fprintf( f, "\n/* end of dttab.cpp */\n");
 	fclose(f);
-	char temp[MAX_PATH];
+	char temp[CSE_MAX_PATH];
 	xfjoinpath(cFilesDir, "dttab.cpp", temp);
 	update( temp, buf);                         // compare file, replace if different
 
@@ -1376,7 +1376,7 @@ LOCAL void wUnits(              // write units info to units.h if different
 LOCAL void wUntab()                     // write untab.cpp
 {
 // open working file untab.cx
-	char buf[MAX_PATH];
+	char buf[CSE_MAX_PATH];
 	xfjoinpath(cFilesDir, "untab.cx", buf);		// buf also used to close
 	FILE* f = fopen( buf, "w");
 	if (f==NULL)
@@ -1811,7 +1811,7 @@ LOCAL RC recs(                  // do records
 	MaxNfields = 0;                     // max # fields in a record
 
 	// open and start "small record & field descriptor" output .cpp file
-	char fsrfdName[_MAX_PATH]; // pathname.cx for small frd output file
+	char fsrfdName[CSE_MAX_PATH]; // pathname.cx for small frd output file
 	FILE* fSrfd = NULL; // small frd output FILE if CFILESOUT, else NULL
 	if (CFILESOUT)                                      // if outputting tables to compile & link
 	{
--- a/src/RCDEF/rcdef.cpp
+++ b/src/RCDEF/rcdef.cpp
@@ -1436,7 +1436,7 @@ LOCAL void wUntab()                     // write untab.cpp
 	/* terminate file, close, update if different */
 	fprintf( f, "\n/* end of untab.cpp */\n");
 	fclose(f);
-	char temp[MAX_PATH];
+	char temp[CSE_MAX_PATH];
 	xfjoinpath(cFilesDir, "untab.cpp", temp);
 	update( temp, buf);                 // compare, replace if different
 }                       // wUntab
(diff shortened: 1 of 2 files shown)


src/RCDEF/rcdef.cpp
```diff
@@ -263,7 +263,7 @@ struct SWTABLE	// terminate w/ last array entry of NULL, default/not found indic
 	int val;
 };
 //-----------------------------------------------------------------------------
-int looksw(			// string/word table lookup, case insensitive
+int looksw_rc(			// string/word table lookup, case insensitive
```
nealkruis, line 266: What is the difference between this function and the other function with the same name?

src/cnglob.h
```diff
@@ -16,6 +16,8 @@
 #pragma warning( disable: 4996)			// do not warn on ISO deprecated functions (stricmp, ) ?C9?
 #pragma warning( disable: 4244 4305)	// do not warn on double->float conversion
 #pragma warning( disable: 4065)			// do not warn if only 'default' in switch
+#else
+#define _countof(array) (sizeof(array) / sizeof(array[0]))  // Defines MSC macro
```
nealkruis, line 20: Not mentioned in the PR description.

src/RCDEF/rcdef.cpp
```diff
@@ -3240,7 +3239,10 @@ LOCAL int gtoks(                 // Retrieve tokens from input stream "Fpm" acco
 						{
 							while ((c=fgetc(Fpm)) != '*')     // ignore til *
 								if (c == EOF)
-									goto ueof;
+								{
+									rcderr("Unexpected EOF: missing '*END'?");
+									return (gtokRet = GTEOF);
+								}
```
nealkruis, line 3245: This is not very DRY.

src/RCDEF/rcdef.cpp
```diff
@@ -3300,7 +3302,10 @@ LOCAL int gtoks(                 // Retrieve tokens from input stream "Fpm" acco
 			{
 				int nchar = fscanf(Fpm,"%s",token);
 				if (nchar < 0)
-					goto ueof;
+					{
+						rcderr("Unexpected EOF: missing '*END'?");
+						return (gtokRet = GTEOF);
+					}
```
nealkruis, line 3308: Ditto.

nealkruis (changes requested):

src/RCDEF/rcdef.cpp
galanCA: No difference at all. Here are the two locations [rcdef.cpp/looksw](https://github.com/cse-sim/cse/blob/f93628bf1240adf3d032dd9d270aa6de706f4f04/src/RCDEF/rcdef.cpp#L266), [lookup.cpp/looksw](https://github.com/cse-sim/cse/blob/f93628bf1240adf3d032dd9d270aa6de706f4f04/src/lookup.cpp#L80). I believe the intent was to remove `lookup.h` as a dependency of rcdef.

src/cnglob.h
galanCA: Added.

src/RCDEF/rcdef.cpp
galanCA: Maybe I can create an inline function that contains the above code. Another possibility is instead of jumping back, we can jump forward after the last current return. My preferred method will be to create an inline function.

src/RCDEF/rcdef.cpp
galanCA: See the above comment.

src/RCDEF/rcdef.cpp
nealkruis: But it is a dependency, hence the redefinition error. You are making decisions here without understanding the full context. In this case, you should ask if we really need separate definitions and get a definitive answer. I am not sure if there is any value in keeping them both, so I will defer to @chipbarnaby. I would imagine we can remove the definition in rcdef.cpp.

src/RCDEF/rcdef.cpp
nealkruis: Give it a try!

nealkruis (changes requested):
See responses above.

src/RCDEF/rcdef.cpp
chipbarnaby: I don't recall the full context either.  I thought I got rid of the lookup.cpp dependency by duplicating this little function.  It worked on MSVC, but I guess not on GCC.  If it is a duplicate definition (without the _rc), then lookup.cpp must still be linked (?).

In general, I don't think it is worth a huge amount of work to achieve absolutely pure non-DRYness, especially in an internal tool like RCDEF.  It is obviously a matter of judgement, but at some point, just do what is needed to make it work, IMHO.

Maybe I'll try to build this on MSVC to understand the issue.

src/RCDEF/rcdef.cpp
chipbarnaby: Is this just to avoid a goto?  There are lots of gotos in CSE.  I would leave them alone unless you want to take the time to set up test cases and verify the changes work.  That kind of testing is time consuming because you have to construct cases that cause all the variants of the error.  "If it ain't broke ..."

src/RCDEF/rcdef.cpp
nealkruis: I believe it is linked through messages and xiopak.

src/RCDEF/rcdef.cpp
galanCA: Try which of those options? I tried both before proposing them.

chipbarnaby (changes requested):
The description has not been updated.  The looksw_rc() rename is no longer included.

Other than that, it looks fine.

nealkruis (approved):

PR #371, opened 2023-01-21
Fix report file comment line

galanCA:
## Description

- Replace faulty strReplace() with safe version that checks destination buffer size.
- Updated strReplace() unit tests.
- Several strpak.cpp and calling code cleanups.
- Use strReplace() to add the prefix option to command line for display in output

## Test
- [x] Add new unit test 

## Author Progress Checklist:

- [x] Open draft pull request
    - [x] Link the issue(s) addressed by this PR (under "Development" in the sidebar menu) 
- [ ] Self-review of code
    - [x] I have made corresponding changes to the documentation
    - [x] Add any unit test for proof and documentation.

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve

### DIFF
--- a/src/cgresult.cpp
+++ b/src/cgresult.cpp
@@ -1772,7 +1772,7 @@ LOCAL void FC vpUdtRpColHeads( DVRI *dvrip)		// user-defined report column heads
 
 		// now store leading blanks, text, 1 trailing blank only
 		while (s < p)  *s++ = ' ';			// rest of blanks before this column
-		memcpyPass( (void **)&s, text, acWid);		// store text / point past
+		memcpyPass( s, text, acWid);		// store text / point past
 		if (i)						// separator char, if not last column (i is colp->nxColi):
 			if (s < p + colWid || RcolB.p[i].colGap)	// not if this col full and next col has 0 gap
 				*s++ = ' ';   				//  one space now, rest when size next col's datum known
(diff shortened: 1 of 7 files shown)


chipbarnaby (changes requested):
I would not solve this problem this way.  The problem is that this approach modifies tp_cmdLineArgs.  Downstream uses of tp_cmdLineArgs (present and future) will have to contend with intermingled prefixes.  Further, if this approach were taken, the comment on tp_cmdLineArgs should be modified to note that the string could have been prefixed-ized.

The approach I would take is to modify the string at the time of display.  The prefix thing is a display feature, so that is where is should be applied -- i.e. inside vrPrint().  I will look at how that works and make a suggestion in an edit of this comment.

galanCA:
@nealkruis Did we mean to update the documentation in this branch?

nealkruis:
@galanCA there are no changes here that affect the documentation. We talked about removing the character limit on `TOP repCpl` and clarifying how it relates to `REPORT rpCpl`. We can do that here, or in another branch. It would be nice to do it here.

nealkruis (approved):
@chipbarnaby Ready for you to take a quick look. @galanCA identified a bug in the `strReplace` function that is also addressed here. Effectively, `iStart` in the function was not moving to the location of the next instance of `sOld`.

chipbarnaby (changes requested):
Comments on cse.cpp --

1. This sequence leaks memory.  strSave() does a heap allocation, but there is no free.  The temporary strpak functions should be used.  They allocate from a self-deleting temporary buffer.  Temporary strings persist for "a while", certainly enough for this use.  The strtcat result can be used directly (without strSave()).  For pfxCmdLineArgs, either use a stack array (and strcpy into it) or a strTemp buffer (but see next comment).
2. strReplace() replaces in place.  In this case, that means the string will get longer.  No extra space has been allocated for this expansion, so as coded, this can write beyond the end of the buffer.
3. It is weird that the initial prefix is provided via the vrPrintf() formatting, while the others are handled using strReplace().
4. I think the prefix-ifier should be made into an strpak function, given that other long strings may have to be written with the prefix.  Something like --

char* strPrefixMultiLineString( char* res, size_t resSize, const char* s, const char* pfx);


Comments on strReplace() --

1. Good that you found the bug.
2. The current algorithm can copy portions of the string several times.  If we dropped the "in place" scheme and had the caller provide space for the result string, then there could be a once-through  copy sequence.  Also, the function could be made safe by passing the size of the output buffer.  Perhaps --

char* strReplace( char* res, size_t resSize, const char* s, const char* sOld, const char* sNew, bool bCaseSens=true)

Return value = res as a convenience.  Or we could retain count as the return value.

Also take note of possibly useful strpak functions, such as memcpyPass()

The only current uses of strReplace() are in strRemoveCRLF() (plus the new use just added),  strRemoveCRLF() is not used, so it could be removed.  Or change it to use strReplace2(), since it is changing single chars.


General comment -- some of the strpak functions are poorly commented.  For example, strReplace() does not document its return value.  When working on functions like this, take the time to correct and/or improve comments.  Leave the code better than you found it.

chipbarnaby (approved):

nealkruis (approved):

PR #403, opened 2023-06-12
Flowmult

chipbarnaby:
## Description

Initial implementation of IZXFER flow multiplier scheme to support zone multiplier modeling approach.

Only DOAS air source is implemented.  Interzone flow multipliers should be considered and added if needed.

Stub documentation provided.


src/cgcomp.cpp
```diff
@@ -1159,29 +1148,26 @@ RC DOAS::oa_BegSubhr()
 {
 	RC rc = RCOK;
 	// Get total air flow rates and calculate mixed exhaust air state
-	oa_supAF.af_amf = 0.;
-	oa_exhAF.af_amf = 0.;
+	oa_supAF.af_Init();
+	oa_exhAF.af_Init();
 	float supVf = 0.f;
-	float exhVf = 0.f;
+	// float exhVf = 0.f;  not used
```
nealkruis, line 1154: This should just be removed instead of commented.

doc/src/records/izxfer.md
```diff
@@ -103,6 +103,17 @@ Name of DOAS where air is supplied from (**izVfMin** > 0), or exhausting to (**i
   required: "when izNVType = AIRNETDOAS",
   variability: "constant") %>
 
+**izLinkedFlowMult=*float***
+
+Flow multiplier TODO
```
nealkruis, line 108: This is not complete.

src/cgcomp.cpp
```diff
@@ -1159,29 +1148,26 @@ RC DOAS::oa_BegSubhr()
...
-			}
+			supVf += izx->iz_vfMin * izx->iz_linkedFlowMult;
+		}
+		else
+		{
+			ZNR* zpx = ZrB.GetAt( izx->iz_zi1);
+			AIRSTATE zoneAir;
+			zpx->zn_GetAirStateLs( zoneAir);
+			float rhoX = zpx->zn_rho0ls;
+			// exhVf -= izx->iz_vfMin * izx->iz_linkedFlowMult;
```
nealkruis, line 1168: ditto

nealkruis (changes requested):
The PR title also needs to be more descriptive.

src/cgcomp.cpp
chipbarnaby: Expanded comments and retained.

nealkruis (approved):

PR #404, opened 2023-06-12
Record info internals: add struct SFIR member to identify fields requiring special case (non-bitwise) copy

chipbarnaby:
## Description

Internal change to struct SFIR.  Added member fi_nxsc to allow a linked list of fields requiring non-bitwise processing during e.g copy.  Altered RCDEF to populate fi_nxsc.

Motivation: provides support for generalized record::Copy().  This will allow elimination of semi-redundant record::Copy() and record::CopyFrom().  Plus eliminate many overrides of record virtuals.

Also added fi_ prefix to SFIR members.

No results changes expected.

### DIFF
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -4,7 +4,7 @@ runs:
   using: composite
   steps:
     - name: Install pandoc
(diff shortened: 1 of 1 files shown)


nealkruis (changes requested):
PR title needs to be more descriptive and `#if 0` code should be removed or commented better.

PR #406, opened 2023-07-10
Support double-precision data in reports; experiment with double-precision MTR data

chipbarnaby:
## Description

Modifications to built-in report generation machinery in cgresult.cpp provide support for double-precision data.  The report scheme is table-driven with one row per report column.  A new flag allows indicating that the data being accessed is double as opposed to float.

Improved generalized data access functions added to support that change.

Testing was performed by changing MTR end use data via conditional code controlled by METER_DBL (cndefns.h).  Reported results were virtually identical to those with float data even for complex MTR schemes with submeters.  METER_DBL left #undef'd but the conditional code retained for future experiments.

Test file submeter.cse updated to better test submeters with multipliers.

### DIFF
--- a/src/CNRECS.DEF
+++ b/src/CNRECS.DEF
@@ -5441,8 +5441,45 @@ RECORD MTR_IVL "meter interval sub" *SUBSTRUCT  // substruct for one meter for o
 	*declare "void mtr_Accum1( const MTR_IVL* mtrSub1, IVLCH ivl, int options=0);"
     *declare "void mtr_AccumFromSubmeter( const MTR_IVL* subMtr, float mult);"
 	*declare "double mtr_NetBldgLoad() const;"
-	*declare "RC mtr_Validate( const MTR* mtr, IVLCH ivl) const;"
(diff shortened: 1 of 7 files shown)


src/CNRECS.DEF
```diff
@@ -5441,8 +5441,45 @@ RECORD MTR_IVL "meter interval sub" *SUBSTRUCT  // substruct for one meter for o
...
+    *e dbl dish		// dish washing
+    *e dbl dry		// clothes drying
+    *e dbl wash		// clothes washing
+    *e dbl cook		// cooking
+    *e dbl usr1		// user-defined end use 1
+    *e dbl usr2		// user-defined end use 2
+	*p dbl bt			// battery output (negative)
+    *e dbl pv			// photovoltaic array output (negative)
+	*e dbl allEU		// subtotal, clg .. usr2 (= load w/o bt and pv)
+#else
```
nealkruis, line 5482: Do we want to keep Float around? Is there any reason not to remove it?

nealkruis (approved):

PR #409, opened 2023-07-17
Alternative string pointers (preliminary support for 64-bit)

chipbarnaby:
## Description

To support a 64 bit implementation, the PR replaces char* pointers within records with 4 byte handles that allow access to strings.  This allows 4 byte NANDLE values to be used to link to string expressions.  On 64 bit systems, char* pointers are 8 bytes while floats are 4 bytes which would complicate the NANDLE/expression scheme.

A new test "commands.cse" has been added that exercises some tricky cases related to object manipulation and string expressions.

No changes in input or other function.  All tests pass with small-if-any results changes.


.gitignore
```diff
@@ -12,6 +12,8 @@
 /test/error_handling/*.spl
 # !/test/*.exe see below
 
+test2/
```
nealkruis, line 15: This is strange, but mostly harmless.

src/cnguts.cpp
```diff
@@ -918,15 +918,19 @@ RC FC cgRddInit(	// Perform initialization common to main simulation run and eac
...
+
 // Initialize for location
 	CSE_EF( Top.tp_LocInit() )	// calls slinit() w wthr file location data (via Top).
 								// inits Locsolar (comments below)
 								// CSE_EF: return RCBAD on any non-RCOK return.
 
-// Set up weather file, TDV file, and DESCONDs
-//   Do each phase, allowing file or location changes.
-//   Messages errors.
-	CSE_EF( Top.tp_WthrInit() )	// also uses Top .tp_wfName, .skyModel [,.isDT]
```
nealkruis, line 929: Was this partly redundant?

src/cvpak.cpp
```diff
@@ -311,7 +311,25 @@ p		break;
 
 	case DTFLOAT:
 floatCase:				// number-choice comes here (from default) if does not contain choice
-
+#if 1
```
nealkruis, line 314: Can we remove the `#if 1`?

nealkruis (approved):

src/cnguts.cpp
chipbarnaby: Yes.  The function actually did not do what the comments said.

src/cvpak.cpp
chipbarnaby: Will look after merge.

PR #413, opened 2023-08-04
Eliminate circular DTYPES dependency in RCDEF

chipbarnaby:
## Description

Rework build scheme so rcdef does not use dtypes.h, eliminating circular dependency.  This allows straightforward switch between 32 and 64 bit builds.

Files included in the rcdef build have been modified to compile using only a small subset of dtypes.h definitions.  Those are provided in cnglob.h when NODTYPES is #defined.  Some dtypes.h typedef types have been coded out in non-cnrecs.def usage (notably BOO (= short integer) has been replaced by bool).

Also combined build and test jobs in the CI workflow.  This allows unit tests to use the current dtypes.h.


CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
+cmake_minimum_required(VERSION 3.15.0)
```
nealkruis, line 1: Is this a necessary update?

src/libstubs.cpp
```diff
@@ -5,11 +5,13 @@
 // Library stubs -- function definition stubs when not using CSE definitions
 //                  e.g., for RCDEF and unit tests
 
+#if 0
 #include "cnglob.h"
 #include "vrpak.h"
 
 VROUTINFO5 PriRep = { { 0 } };
 const char* InputFilePathNoExt = NULL;
+#endif
```
nealkruis, line 14: Do we need this or not? If it is sticking around, should have some comments about why.

test/unit/CMakeLists.txt
```diff
@@ -3,23 +3,21 @@ set(source
       strpak.unit.cpp
       xiopak.unit.cpp
       cvpak.unit.cpp
-      $<TARGET_OBJECTS:cse_libstubs>
+      ../../src/libstubs.cpp
```
nealkruis, line 6: Use `${CSE_SOURCE_DIR}` instead of `../../src`.

nealkruis (changes requested):

CMakeLists.txt
chipbarnaby: CMake stuff I've read recommends putting cmake-minimum_required as the first thing in the file, so I moved it.

src/libstubs.cpp
chipbarnaby: I agree it should be deleted, will do that.

test/unit/CMakeLists.txt
chipbarnaby: OK

CMakeLists.txt
nealkruis: But no particular reason to upgrade to 3.15 from 3.10?

chipbarnaby: Actually the change was inadvertent.  I have done some reading in Professional CMake and noted many features added in 3.15 related to debugging, clang-on-Windows-support, and other things.  Given that 3.15 is now 4 years old, it is not unreasonable to require it.  In fact, I would advocate going much newer and learning how to exploit current features.  We're a small group, it would not be a major inconvenience to move to a newer version.

PR #423, opened 2023-08-29
Improved field number type checking; correct bad usage detected

chipbarnaby:
## Description

64-bit development found an instance of IsSet( record-member) where IsSet( record-member-field-number) was intended.

This PR includes several changes to field status access and testing functions to check that the fn (field number) is an integer.  With this, the original offending error was detected along with several more.  All have been fixed.

A more robust fix would be to define a type for field number.  An issue has been added.

No results changes.  No regression test changes.  No documentation changes.

### DIFF
--- a/src/ancrec.h
+++ b/src/ancrec.h
@@ -48,18 +48,6 @@ typedef basAnc* BP;	// basAnc pointer -- formerly used to localize NEARness
 	         re field-interdependent error messages and defaults.
 	      3. may be tested by cuprobe.:probe() re input time probes. */
 
-#if 0
-// unsuccessful experiment re general fn type check, 8-2023
-template <typename T> struct FLDNUM
-{	FLDNUM(T _fn)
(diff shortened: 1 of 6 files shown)


src/ancrec.h
```diff
@@ -48,6 +48,18 @@ typedef basAnc* BP;	// basAnc pointer -- formerly used to localize NEARness
...
+template <typename T> struct FLDNUM
+{	FLDNUM(T _fn)
+	{
+		static_assert(std::is_same<int, T>::value);
+		fn = _fn;
+	}
+	int fn;
+};
+#endif
+
```
nealkruis, line 62: Doest this need to be kept? If it has promise, maybe it can be moved to an issue instead?

chipbarnaby: Agree, I will delete.

nealkruis (approved):

PR #454, opened 2023-12-18
Replace #include <math.h> with #include <cmath>

chipbarnaby:
## Description

Motivation: standardize math-related declarations using cmath.

Prior to this change, cnglob.h included math.h.  Some vendor headers included cmath.  The Microsoft library can end up using different implementations of some functions (e.g. pow()) depending on include order.  CSE results change when these alternative implementations are selected.  Overall upshot -- adding or re-ordering includes of vendor headers could cause apparently unrelated results changes.

The primary code change in this PR is changing cnglob.h to #include \<cmath\>.   Some redundant includes were removed.  Related comments are updated.

Many regression tests show differences as a result of this change.  Most are deemed trivial.  In all cases, annual results are virtually unchanged. Larger and potentially significant changes are seen in hourly results in the following --
- ashpvc2
- chdhw
- submeter
- MF8X


src/cgcomp.cpp
```diff
@@ -2283,8 +2276,6 @@ void TOPRAT::tp_AirNetDestroy()
 //   finds zone pressures that achieve balanced mass flows
 ///////////////////////////////////////////////////////////////////////////////
 #if defined( AIRNET_EIGEN)
-// Eigen #include located here because placement of top of file altered
-//   code generation associated with pow() (4-12-2023).  Cause not known.
 #include <Eigen\Dense>
```
nealkruis, line 2279: Move this up now?

chipbarnaby:
I decided to leave it where it is (within the #if defined( AIRNET_EIGEN))

src/cgcomp.cpp
chipbarnaby: I decided to leave it within the #if

PR #460, opened 2024-01-19
Btwxt error handling

chipbarnaby:
## Description

Motivation: capture and report btwxt error messages.

Implemented several classes derived from Courier.  Updated all btwxt uses.


nealkruis:
Here's how I would design this:

In `hvac.h`:
```c++
class RSYSBtwxtLogger : public Courierr::Courierr
{
public:
	explicit RSYSBtwxtLogger(std::string rgi_name_in, RSYS* pRSYS_in) : rgi_name(rgi_name_in), pRSYS(pRSYS_in) {}
	void error(const std::string_view message) override { pRSYS->oer(format_message(message)); }
	void warning(const std::string_view message) override { pRSYS->oWarn(format_message(message)); }
	void info(const std::string_view message) override { pRSYS->oInfo(format_message(message)); }
	void debug(const std::string_view message) override {  pRSYS->oInfo(format_message(message));  }
private:
	RSYS* pRSYS;
	std::string rgi_name;
	std::string format_message(std::string message) { return fmt::format("btwxt '{}' -- {}", rgi_name, message)}
}
```

and I'd make a similar one for CHDW. You could even put `rgi_name` and `format_message` into a baseclass that these are both derived from. That way you don't have to deal with the callbacks and casting.

src/hvac.h
```diff
@@ -15,14 +15,83 @@ float ASHPCap95FromCap47( float cap47, bool useRatio, float ratio9547);
...
+		else
+			err(PABT, "nullptr cmh_pMsgCallbackFunc '%s'", msg);
+
+	}
+
+private:
+	void* cmh_context;						// caller context
+	MsgCallbackFunc* cmh_pMsgCallbackFunc;	// pointer to callback function
+
+};	// class CourierMsgHandler
```
nealkruis, line 63: My thought is that `void* context` and a `MsgCallbackFunc*` are very "C". There is a lot of indirection for people trying to read the code. My preference would be to make a new class for each different context with a specific `forward_message` override.

src/strpak.cpp
```diff
@@ -391,6 +392,14 @@ char* strncpy0(			// copy/truncate with 0 fill
...
 }		// ::strncpy0
 // ====================================================================
+char* strt_string_view(		// copy string_view to temporary
+	std::string_view sv)	// string view
+{
+	// strncpy0 handles sv.size() = 0
+	char* sRet = strncpy0( nullptr, sv.data(), sv.size()+1);
+	return sRet;
+}		// :: strt_string_view
+// ====================================================================
```
nealkruis, line 402: Is this still needed?

src/strpak.h
```diff
@@ -92,6 +92,7 @@ char * FC strtempPop( char *anS);
 char * FC strtmp( const char *s);
 char * CDEC strtcat( const char *s, ... );
 char * CDEC strntcat( int n, ...);
+char* strt_string_view(std::string_view sv);
```
nealkruis, line 95: see previous question.

nealkruis (commented):
Consider changes. I won't push too hard on my suggestions if you want to merge anyway.

src/strpak.cpp
chipbarnaby: Not right now, but I made it and tested it.  It is tiny and likely useful soon given the ever-growing interaction with other libraries.  So I say leave it.

src/hvac.h
chipbarnaby: 1. Don't want 10 essentially identical implementations.  Maybe there is a way to template this to make it clearer?  In any case, there is going to be a "callable" somewhere aka a pointer.
2. Additionally, what is wrong with C?  CSE is small and fast, partially due to using lots of C stuff.
3. I recently learned about a "dependency injection" scheme which might be very useful here.

Issue #465, opened 2024-02-07
String concatenation


chipbarnaby:
Add a string concatenation function to the CSE language

str3 = concat( str1, str2)

Useful for e.g. assembling case-specific file names or informative report titles.

A preliminary implementation code appears to work correctly but has not been sufficiently tested or generalized, see PR #

In addition to testing, the following are worth considering:

- Make the function take n arguments (now limited to 2).
- Use a binary operator ("+" or similar) instead of a function.

PR #466, opened 2024-02-07
Preliminary string concatenation function

chipbarnaby:
## Description

Add to CSE language a preliminary version of a string concatenation function.  See issue #465.

concat() function is disabled in this commit.  Code is a placeholder for further development.

Also included are independent changes to .gitignore to reduce git status "untracked" clutter.


.gitignore
```diff
@@ -6,20 +6,26 @@
 /test/*.csv
 /test/*.err
 /test/*.spl
+/test/*.bat
+/test/Testing/
```
nealkruis, line 10: This shouldn't be here at all. A `Testing` directory indicates that you ran `ctest` from the wrong directory. It shouldn't be ignored because it should be obvious when someone makes this mistake, and they should delete the resulting directory.

.gitignore
```diff
@@ -6,20 +6,26 @@
...
+/test/ref64/
```
nealkruis, line 11: I don't think we want to ignore this either. We likely want to commit it at some point.

.gitignore
```diff
@@ -6,20 +6,26 @@
...
 /test/error_handling/*.rep
 /test/error_handling/*.csv
 /test/error_handling/*.err
 /test/error_handling/*.spl
 # !/test/*.exe see below
 
+projects/
 test2/
+sandbox/
+src/attic
```
nealkruis, line 21: This one seems pretty specific...

src/RCDEF/rcdef_command_line.txt
nealkruis, line 1: What is the purpose of this?

nealkruis (changes requested):

nealkruis (approved):

PR #473, opened 2024-03-19
Avoid infinite loop

spahrenk:
## Description
- Prevents simulation hang-up that resulted from `nan` in `cvpak` function `cvDouble2s()`.
- Allows simulation to proceed in case of `secant` failure in `RSYS::rs_AllocateZoneAir()`.
- Expands range of solution by defining a reduced tolerance used in `ZNR::zn_AmfHvacCR` and `ZNR::zn_AmfHvacCR`.
- Repeats, while reporting, iteration steps in `secant` in case of failure.
- Adds unit test for `nummeth`, specifically involving `secant`.

Note that failure in `RSYS::rs_AllocateZoneAir()` can still occur, when the solution for available supply air temperature approaches within the defined tolerance of the zone air temperatures for any of the served zones. But in the case of such failure, the previous supply air-temperature value is used, allowing the simulation to proceed.


chipbarnaby:
Shouldn't the nan case be intercepted much earlier?  Probably in cvin2s()?  If the data is double or float and isnan(), emit "nan".  There is probably a need to worry about field width.

There already is a cvin2s unit test.  First step might be to add one or more "nan" cases so we can easily verify what happens.

src/cvpak.cpp
```diff
@@ -528,8 +528,9 @@ p       Cvnchars = sprintf( str, sif[lj][ipv], wid, ppos, 0);
 		for (i = 0; ; i++)
 		{
 			// more robust to give up HERE if i too big ?? 10-88 rob
-			if ( fabs(val) < maxfit		// if now might fit width
-			&&  cvdd( mfw-1, dfw) )  		// format it and see
+			if ((fabs(val) < maxfit		// if now might fit width
+			&&  cvdd( mfw-1, dfw)) ||  		// format it and see
+				std::isnan(val))            // guard against inf loop
```
chipbarnaby, line 533: I stand by my prior comment that the nan case should be intercepted earlier.  Suitable output should be generated, e.g. "nan".  Maybe issue warning?  (Or maybe not ... printf() does not.  What does fmt do?)   Also, as suggested add one or more nan cases to the cvpak test to exercise the new capability.

src/nummeth.cpp
```diff
@@ -235,7 +235,89 @@ int secant(							// find x given f(x) (secant method)
     }
 	return i;
 }			// ::secant
+
 //-----------------------------------------------------------------------------
+int secant( // screen secant success; report calcuation if failure
```
chipbarnaby, line 240: I don't understand the strategy here.  There should be high-level explanatory comments.  "Try classic secant method.  If it fails converge, do x, y, z in an effort to allow execution to continue."

chipbarnaby (commented):
See individual comments,

src/cnglob.h
```diff
@@ -271,6 +271,7 @@ constexpr double tAbs0F						// 0 F in Rankine
 #else
 	= 459.67;
 #endif
+constexpr double tol_tF = 1.e-12; // temperature tolerance (degF)
```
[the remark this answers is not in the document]
spahrenk: The tolerance used in these two cases serves to prevent division by zero; it does not correspond to a precision on the fitted value.  (The fitting tolerance and number of iterations within the secant method have not been altered.) Even with the tighter tolerance, the `secant' method will likely fail if the if the condition  in these functions is met. (It seems obvious that the the two functions are related, but if there is some justification for using a different value in one or the other, please specify). Given that there are at least two instances of solver functions that return AMFs and contain terms with degF temperature differences  in their divisors, the definition of a standard tolerance within some broader scope seems merited.

src/nummeth.cpp
spahrenk: I was told to add code to repeat `secant` upon failure with text output. I have expanded some comments to clarify.

nealkruis (approved):

PR #478, opened 2024-04-15
Add generic models hpwh

spahrenk:
## Description

Enables use of Tier-4 HPWH models in cse.


nealkruis:
@spahrenk, HPWHsim "main" is on 2c3b199; but the HPWHsim submodule on this branch is still pointing to b9efd3b.

Once that is updated, I think this one will be all set pending @chipbarnaby's review (which can happen in parallel since there are no other API changes expected).

chipbarnaby (approved):
The 0.0000 -> -0.0000 changes are Interesting but not worth pursuing.

PR #482, opened 2024-04-26
Add generic entries.

spahrenk:
## Description
Added descriptions of generic models to user manual, as provided by B. Larson.


nealkruis (approved):
Thanks, @spahrenk!

PR #499, opened 2024-08-29
Update kiva/reduce boost

spahrenk:
## Description

Update kiva submodule. New version includes boost-1.86.0 reduced using the bcp utility.


nealkruis:
@chipbarnaby the failure on mac build here is related to the "W" suffix version of `isdigit`, when third party code is using the original versions of the functions. Is there a cleaner solution where we don't have to re-define common std functions?

chipbarnaby:
I'm sure we can find a better solution.  I believe the issue relates to
signed chars being promoted to negative integers.

PR #504, opened 2024-09-06
Eliminate strpak.h charTypeErr scheme

chipbarnaby:
## Description

Eliminate redefinition of ctype functions (e.g. isspace) to cause a compiler error.  This scheme allowed detection of possibly incorrect use of such functions.  However, downstream 3rd party #include files that use ctype functions could not be compiled.  In particular, the Mac build failed due to this approach.

Also changed #include <ctype.h> to #include <cctype> as in incremental step toward use of C++ headers as opposed to legacy C headers.

### DIFF
--- a/src/cnglob.h
+++ b/src/cnglob.h
@@ -84,7 +84,7 @@ struct CULT;
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <ctype.h>
+// #include <cctype> .. in strpak.h
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


nealkruis:
@chipbarnaby, have you confirmed that the "W" variants are still needed in some cases? Doesn't have to hold this up, but if you have maybe we can get rid of them all together? I was never able to reproduce the issue described in these comments.

nealkruis (approved):

PR #516, opened 2024-11-13
Add AOSmith and LG HPWH models.

spahrenk:
Adds AOSmithHPTS40 and LG_APHWC50 HPWH models. Documentation was expanded to include these (and new Bradford White) models.


nealkruis:
We need to sort out issues with performance extrapolation before we can merge this PR.

PR #532, opened 2025-02-26
Fix large-draw bug (HPWHsim)

spahrenk:
Modifies HPWHsim to fix flawed calculation when the draw volume exceeds the tank volume. The case Vdraw > Vtank  was formerly handled with separate code, which simplified the calculation by ignoring possible temperatures inversions in the tank. However, inversions are only to be expected when Tinlet > Ttank[0] (i.e., the bottom tank node), which is uncommon. (Perhaps the alternative code was introduced to improve computational speed.) Furthermore, separate algorithms for high and low draws can lead to discontinuities in the simulation results when Vdraw ~ Vtank. This PR applies the same algorithm (including possible inversion mixing) , regardless of draw volume.

The large-draw condition appears not to have been previously tested in HPWHsim. (Bug appear to have been introduced in [balance-energy-hpwh](https://github.com/cse-sim/cse/tree/balance-energy-hpwh)) This PR adds energy-balance tests to check this case.


chipbarnaby (approved):
Looks plausible to me, since I benefit from not understanding to code in detail.

PR #536, opened 2025-03-27
Add heating charge factor

chipbarnaby:
## Description

Generalized RSYS charge factor scheme.
- Added rsFChgH = heating compressor charge factor.  Implemented in both "old" and PM models.
- Renamed rsFChg to rsFChgC = cooling compressor charge factor.  (rsFChg is supported for backwards compatibility.)
- Modified implementation of charge factors to increase COP (or input power) rather than decrease capacity.

These changes results in smaller autosized capacities, but that is expected and has only a minor impact on energy results.

ALSO merged branch "modernize" that allows mapping of old to new probe names.

Small test results changes (due to autosizing differences).  Reference files are reviewed and updated.

Documentation updated.


src/srd.h
```diff
@@ -191,6 +190,39 @@ struct VALNDT
...
+		return mp_oldWord==nullptr;
+	}
+	bool mp_ModernizeIf(
+		const char* &word) const	// word that may need modernizing
+									//  e.g. from user input
+									// returned updated if needed
+	// returns true iff word modernized (word updated)
+	//    else false
+	{
+		bool bMatch = _strcmpi(word, mp_oldWord) == 0;
```
nealkruis, line 218: `_strcmpi` is not standard C++

chipbarnaby:
You could fix this to keep things moving.

nealkruis (approved):

PR #542, opened 2025-05-02
First 10 airnet converge failures treated as warnings

chipbarnaby:
## Description

Change handling of airnet convergence failures to tolerate a few convergence failures.  Convergence failures have been seen occasionally in models with a large number of zones (e.g. multifamily cases with e.g. 80 zones).

Was: Max 20 iterations, 1st failure = error (run terminated)

Now: Max 30 iterations, 10 failures treated as warnings (run continues).  11th failure is error.

No results changes (except for runs that formerly did not complete)

No documentation changes.

### DIFF
--- a/src/dhwcalc.cpp
+++ b/src/dhwcalc.cpp
@@ -2537,6 +2537,10 @@ float DHWSYSRES_IVL::wsr_EnergyBalance()	// calculate energy balance
 {
 	float otherSum = VSum(&qLossMisc, wsr_NFLOAT - 3);
 	qBal = qOutDHW + qOutHtg - otherSum;
+#if defined( _DEBUG)
+	if (fabs(qBal) > 5000.f)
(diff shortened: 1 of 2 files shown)


src/dhwcalc.cpp
```diff
@@ -3662,12 +3666,16 @@ RC HPWHLINK::hw_DoSubhrTick(		// calcs for 1 tick
 {
 	RC rc = RCOK;
 
-#if 0 && defined( _DEBUG)
-	if (Top.tp_date.month == 7
-		&& Top.tp_date.mday == 27
-		&& Top.iHr == 10
+#if 1 && defined( _DEBUG)
```
nealkruis, line 3669: Was this left "on" intentionally?

chipbarnaby: Fixed in PR #541

PR #545, opened 2025-05-14
Update izAFCat documentation

nealkruis:
## Description

Replaces documentation of general interzone airflow categories with conditioned/unconditioned distinctions.


nathancoliver (approved):
The izxfer.html looks good to me.

PR #549, opened 2025-05-30
Add probe for daily min temperature (taDbMin)

chipbarnaby:
## Description

Added tracking of the daily minimum temperature, probable hourly as @weather.taDbMin.

Renamed @weather.taDbPk to @weather.taDbMax so naming is parallel to taDbMin.  Similarly, taDbPvPk is now taDbPvMax.

Modernizations provided for probe renames so pre-existing files run successfully.

Updated test file wthr01.cse to report taDbMin, taDbMax, and taDbPvMax for a few days.

No documentation changes.

No test result changes other than updated wthr01.


nealkruis:
@chipbarnaby are we worried at all about backwards compatibility?

Also, can you update the Mac results? We probably need a better mechanism to update reports on other platforms...

chipbarnaby:
No backward compatibility issue.  I have added a modernization scheme in
cuprobe.cpp that maps old probe names to current names.

PR #550, opened 2025-06-03
Update Windows platform toolset to v143

spahrenk:
Changed the Windows platform toolset specified in `configure.cmake` from "Visual Studio 2019 (v142)" to "Visual Studio 2022 (v143)". I have not identified a specific version (toolset, compiler?), as was listed previously. The output of perfmap3.cse on Win32 changed slightly (an extra "Info" message is displayed.) This version passes the CI, but now  my local Mac gives several 0/-0 diffs in perfmap.cse, compared to the CI (due to AppleClang 14 vs. 15?).

### DIFF
--- a/cmake/configure.cmake
+++ b/cmake/configure.cmake
@@ -16,7 +16,7 @@ endif ()
 
 if ("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
     if (${COMPILER_ID} STREQUAL "msvc")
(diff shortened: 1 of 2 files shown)


nealkruis:
Here's the list of toolset versions: https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

The latest SDK versions can be found here: https://en.wikipedia.org/wiki/Microsoft_Windows_SDK

Not sure which SDK version we should target. Not sure it matters?

PR #551, opened 2025-06-04
Initialize HPWHsim presets from embedded CBOR

spahrenk:
The initialization of HPWH models is changed to load from a `hpwh_data_model` representation embedded within the HPWHsim source code in CBOR format, rather than via separate code-based setup for each model. These representations were themselves generated from the original code-based setups, which were converted to JSON, then embedded as CBOR. These JSON model files are included with the HPWHsim repository at test/model_json. The embedded representations are generated from these JSON files during the CMake build.


.github/workflows/build-and-test.yml
```diff
@@ -53,6 +53,10 @@ jobs:
       - name: Get number of CPU cores # Can replace when CMake is upgraded to 3.29
         uses: SimenB/github-actions-cpu-cores@v2
         id: cpu-cores
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+      - name: Set up Python
+        run: uv python install
```
nealkruis, line 59: Use the same setup as HPWHsim.

test/ref-macos64-appleclang/BGTEST.REP
nealkruis, line 1: Only update report files with meaningful differences (i.e., not lines preceded with a `!`).

test/ref-macos64-appleclang/DHW_MFSIZING.REP
```diff
@@ -418,47 +418,47 @@ Monthly Energy Use, meter "MtrElec"
...
+Sep 1935.2      0      0      0 1886.1      0 49.135      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0
+Oct 2129.1      0      0      0 2078.3      0 50.773      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0
+Nov 2472.6      0      0      0 2419.1  4.288 49.203      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0
+Dec 2888.7      0      0      0 2837.9      0 50.773      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0
 
-Yr   29526      0      0      0  28885 43.129 597.81      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0
+Yr   29547      0      0      0  28908 40.590 597.81      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0      0
```
nealkruis, line 434: Is this an expected change? It's more than a a few decimal places.

test/ref-macos64-appleclang/dhwloop32U.rep
nealkruis, line 1: This file also has bigger than expected differences that need to be justified.

test/ref-macos64-appleclang/dhwx.rep
nealkruis, line 1: This file, too.

test/ref-macos64-appleclang/CHDHW.REP
nealkruis, line 1: What's happening in this file?

nealkruis (changes requested):
A few issues here that still need to be resolve. Also needs to be updated with latest `main`.

spahrenk:
@nealkruis Reduced the executable size by moving model initialization to source file.

chipbarnaby:
I think we can temporarily live with an 8.6 mb exe.  But tuning the number of points and implementing external files should still be a priority.

tanaya-mankad:
There's one place where I left in the new BradfordWhite model names, hoping it won't be too much of a problem since adding the full models is up next.

PR #566, opened 2025-08-14
Accumulator

chipbarnaby:
## Description

Accumulator placeholder


nealkruis:
@chipbarnaby which phase did you decide to use when updating the accumulators?

This is still in DRAFT. Let me know when it's ready to merge.

chipbarnaby:
Initial accumulator implementation parallels METER calc sequence.  It has not yet be tested on tricky cases where expression evaluation order might trip things up.  Hence, the PR is draft.

I plan to add nasty test cases and that might alter the implementation and/or the design.

PR #572, opened 2025-09-15
Remove upper limit on COP in performance maps.

nealkruis:
Certain combinations of SEER2 and EER2 input into the RESNET method result in cooling COPs at low temperatures above 20. This is not unrealistic, and shouldn't produce an error.


chipbarnaby (approved):

chipbarnaby:
PR needs description.  Motivation?

PR #616, opened 2026-04-09
Fix delimiter for the PATH environment variable on non-Windows operating systems

nealkruis:
## Description

The existing code for parsing the PATH string assumed semi-colon delimiters, but unix-based systems use colons. This was discovered when attempting to run CSE within a uv virtual environment on MacOS resulting in a SIGABRT crash (likely due to character array overflow?).

## Reviewer Checklist:

     - [ ] Request changes, or
     - [ ] Approve

### DIFF
--- a/src/xiopak.cpp
+++ b/src/xiopak.cpp
@@ -51,6 +51,12 @@ namespace filesys = std::experimental::filesystem;
 //   Functionality replaced with C lib calls and caller local path searches
 //   Also deleted file directry.cpp containing functions used only by deleted xf functions
 
+#if CSE_OS == CSE_OS_WINDOWS
+static constexpr char PATH_DELIMETER = ';';
+#else
+static constexpr char PATH_DELIMETER = ':';
+#endif
+
 
 /*----------------------- LOCAL FUNCTION DECLARATIONS ---------------------*/
 LOCAL SEC FC xioerr( XFILE *xf);
@@ -660,10 +666,10 @@ bool findFile( 	// non-member function to find file on given path
 			if (!path)						// if NULL given
 				path = getenv("PATH");		//   search environment path
 			char* p = strTrim( NULL, path);	// working copy
-			const char* pToks[ 100];
-			int n = strTokSplit( p, ";", pToks, 100);
+			const char* pToks[ CSE_MAX_PATH];
+			int n = strTokSplit( p, &PATH_DELIMETER, pToks, CSE_MAX_PATH);
 			if (n < 0)
-				n = 100;
+				n = CSE_MAX_PATH;
 			for (int i=0; !found && i<n; i++)
 				found = fileFind1( pToks[ i], fName, fNameFound);
 		}
@@ -694,7 +700,7 @@ void Path::add( 		// add ;-delimited paths to Path object
 
 // determine size
 	size_t lenNow = p ? strlen(p) : 0;			// 0 or length of existing path(s) string
-	size_t sneed = (lenNow + strlen(s) + 2) | 32;  	// space needed: old, new, ';', '\0', | 32 for less reallocation
+	size_t sneed = (lenNow + strlen(s) + 2) | 32;  	// space needed: old, new, PATH_DELIMETER, '\0', | 32 for less reallocation
 	size_t snow = (lenNow+1) | 32;			// now big it must be now (if p != NULL)
 
 // allocate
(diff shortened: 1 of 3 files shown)


src/xiopak.cpp
```diff
@@ -660,10 +666,10 @@ bool findFile( 	// non-member function to find file on given path
 			if (!path)						// if NULL given
 				path = getenv("PATH");		//   search environment path
 			char* p = strTrim( NULL, path);	// working copy
-			const char* pToks[ 100];
-			int n = strTokSplit( p, ";", pToks, 100);
+			const char* pToks[ CSE_MAX_PATH];
```
chipbarnaby: It looks like pToks[] is an array of pointers to each segment of the path.  That is, 100 is the max number of path segments handled, not the segment max length.

I think change back to 100 or perhaps 200.

src/xiopak.cpp
chipbarnaby: Fix comments that include literal ';' ?

src/xiopak.cpp
```diff
@@ -694,7 +700,7 @@ void Path::add( 		// add ;-delimited paths to Path object
 
 // determine size
 	size_t lenNow = p ? strlen(p) : 0;			// 0 or length of existing path(s) string
-	size_t sneed = (lenNow + strlen(s) + 2) | 32;  	// space needed: old, new, ';', '\0', | 32 for less reallocation
+	size_t sneed = (lenNow + strlen(s) + 2) | 32;  	// space needed: old, new, PATH_DELIMETER, '\0', | 32 for less reallocation
```
chipbarnaby: I don't fully understand the logic here, but I have slight worry about long paths.  Windows allows length of the path environment var to be something like 32760.  With size_t etc. this code looks OK (no SIs in sight?), but it should be reviewed.  What are the limits on Mac and Linux?

src/xiopak.cpp
```diff
@@ -51,6 +51,12 @@ namespace filesys = std::experimental::filesystem;
 //   Functionality replaced with C lib calls and caller local path searches
 //   Also deleted file directry.cpp containing functions used only by deleted xf functions
 
+#if CSE_OS == CSE_OS_WINDOWS
+static constexpr char PATH_DELIMETER = ';';
+#else
```
chipbarnaby, line 56: Perhaps Mac and Linux should be explicitly called out, with an #else for a new alternative?  That would produce a compiler error if a new alternative were added, forcing review.

chipbarnaby (changes requested):
Some issues in embedded comments.

This points out that our testing environment does not ever involve non-local files.  Perhaps we should add subfolder cse/test/nonlocalfiles that contains some of referenced files and use them in some test cases.  This path problem would have been found long ago if the file locator capability had been exercised.

src/xiopak.cpp
nealkruis: Fixed, and made the "magic number" more explicit.

src/xiopak.cpp
nealkruis: This one escaped. I cleaned up other comments too that were outdated.

src/xiopak.cpp
nealkruis: I'm not touching this...just the comment.

src/xiopak.cpp
nealkruis: I considered this. There are a few edge cases, but I don't think we'll ever need to handle them.

src/xiopak.cpp
```diff
@@ -719,8 +726,8 @@ void Path::add( 		// add ;-delimited paths to Path object
 // update
 	char *q = nup + lenNow;		// pointer to end of current paths
 	if (lenNow)				// if there were already any paths
-		if (*(q-1) != ';' && *s != ';')	// unless there already is a ';'
-			*q++ = ';';			// supply separating ;
+		if (*(q-1) != PATH_DELIMITER && *s != PATH_DELIMITER)	// unless there already is a PATH_DELIMITER
+			*q++ = PATH_DELIMITER;			// supply separating PATH_DELIMITER
 	strcpy( q, s);			// append new path(s)
 	_strupr(q);				/* upper-case user-entered paths for uniform appearance 2-95
```
chipbarnaby, line 732: Didn't see the _strupr() call on 1st review.  This is trouble for case sensitive file systems.

Proposal 1: omit always.  Who cares about "uniform appearance"?

Proposal 2: include only for Windows

All this points to the need to do some actual testing on Mac (and maybe Linux).  Verify that the path mechanism actually works as opposed to not crashing.

chipbarnaby (changes requested):
Prior comments addressed.

New issue found: gratuitous uppercasing of path ??

src/xiopak.cpp
nealkruis: True this is an issue, but I'd say a larger and separate issue from the path delimiter. I suggest we start a new PR that addresses #618.

chipbarnaby (approved):

PR #622, opened 2026-04-19
Fix memory leak due to EXTAB element re-use

chipbarnaby:
## Description

Rework of exman.cpp code to eliminate memory leak that could occur when an EXTAB (expression table) is reused.  One case when this occurred was when an ALTER modified a field that had a non-constant expression value.  In this case, the pre-alter expression code on the heap was not being freed.

Also limited modernization and cleanup.  Some of which extended into other files.

No results changes.

No documentation changes.


nealkruis:
@chipbarnaby is this ready to review? It's good to merge as far as I'm concerned.

chipbarnaby:
I'm doing some cleanup on exman.cpp, will push Real Soon.

nealkruis (approved):

PR #624, opened 2026-04-21
Import bug fix and rework

chipbarnaby:
## Description

Initial motivation = Bug that mishandled import file reads in runs that used autosize.  Autosize involves multiple partial runs and requires restart of import file processing.  Fixed.

Memory leaks were discovered during above work.  Reworked code to replace internal hand-rolled dynamic arrays with std::vector and some char* strings with std::string.

Added regression test import.cse that uses 2 import files with interleaved reading.  Also tests CSE functions import() and importStr() with column references by name and by position.

No results changes (except addition of new test import.cse).


src/CNRECS.DEF
```diff
@@ -1853,18 +1853,28 @@ RECORD DVRI "dvri" *RAT *hideall	// Date-dependent Virtual Report Info -- at run
     					//  (list headed by Top.drviY, -M, D, H, HS, S updated daily so don't need to search subHourly)
 *END		// DVRI
 //=============================================================================
+RECORD FNRT "fnrt" *SUBSTRUCT   // IMPORTFILE substructure -- fields-by-number table
```
nealkruis, line 1856: What does FNRT stand for?

src/CNRECS.DEF
```diff
@@ -1893,11 +1903,19 @@ RECORD IMPF "importFile" *RAT	// IMPORTFILE input (Impfib) and run (ImpfiB) reco
...
     *declare "bool eorScanned;"	// true iff all fields in record have been scanned
     //fields-by-number info. IFFNM.fnmt[] used to get #'s of named fields at runtime.
-	*declare "struct FNRT* fnrt;" // ptr to heap array of info subscripted by 1-based field number.
-    *r *hide USI fnrtNAl 	// allocated size of .fnrt[]. Max subscript fnrtNAl-1; [0] unused.
+    *declare "std::vector<FNRT> im_fnrt;"    // info subscripted by 1-based field number
+                                             //  [ 0] unused
 *END		// IMPF
 //=============================================================================
-RECORD IFFNM "impFileFldNames" *RAT	// IMPORT FILE FIELD NAMES TABLE (IffnmB) records
+RECORD FNMT "fnmt" *SUBSTRUCT   // IFFNM substructure
```
nealkruis, line 1911: Same here. What does FNMT stand for? At a minimum it should be explicit in the comments. Maybe such succinctness is overkill at the expense of readability?

nealkruis (approved):
Just a few minor questions.

Issue #619, opened 2026-04-16
Weather file path corruption


nealkruis:
We continue to see seemingly random cases with errors when looking for weather files in a path string that appears to be referencing bad memory. We have not been successful in isolating the cases where this occurs beyond that it seems to be specific to Windows builds (both 32-bit and 64-bit).

Big Ladder has encountered these issues when performing large-scale batch parametric simulations through its Modelkit framework, but they do occur with Modelkit generated CSE files when run directly from the command line as well.

chipbarnaby:
Do I understand that this bug is seen *only* when the weather file is not in the current working directory?

In other words, the first area to investigate should be locate-a-file-via-path stuff.

Please confirm and provide any other specifics that have been observed.

nealkruis:
> Do I understand that this bug is seen _only_ when the weather file is not in the current working directory?

That is the only case we've seen it, but doesn't mean it doesn't ever occur in the CWD.

So far this has occurred on three separate Big Ladder employee machines (all different makes).

nathancoliver:
Moving this over from [#440](https://github.com/cse-sim/cse/issues/440). Below are the comments from the other issue.

@nathancoliver:

> ```
> ---------------
> Error: R0100: Weather file '͞��': open failed
> ---------------
> C:\USERS\NATHAN OLIVER\DESKTOP\REPOSITORIES\NEEA-ROOM-HEAT-PUMPS\RUNS\CZ1_MIAMI_FL_NEW\CASES\CT2-SF1-2\INSTANCE.CSE(1600): Info: 
>   S0214: No run due to error(s) above
> ---------------
> ```


@chipbarnaby:

> @nathancoliver can you supply:
> 
>    1. all files required to run CSE (that is, instance.cse and all included
>    / referenced files -- I do not need modelkit input)
>    2. A description of your setup.  Are you relying on the environment path
>    to find the weather file or is it in the same folder as instance.cse?  Is
>    the current working directory the location of instance.cse?  That is, is
>    CSE invoked in the instance.cse folder ("cse-folder\cse instance.cse") or
>    elsewhere ("cse-folder\cse instance-folder\instance.cse")?

[...]

PR #630, opened 2026-05-08
Fix weather file path corruption

chipbarnaby:
## Description

Fixes issue #619.

Origin of error = incorrect CULSTR assignment sequence in cuparse.cpp exOrk().

The incorrect logic was located with a hardcoded trap that detected bad tp_wfName.  A sequence of test files were run in a loop using the 64 bit release until the error manifested (usually within 5 minutes).  With the fix, the test loop has been run for several hours without error.

The reason that the problem rarely occurred has not been figured out.  The problem case was never duplicated while debugging, so detailed research was not possible.  Something to do with operating system memory management timing (CULSTR uses the heap for string storage) (?).

Also did some cleanup, especially in CULSTR (strpak.cpp).  Added CULSTR move assignment operator.

No results changes.

No documentation changes.

### DIFF
--- a/src/cgwthr.cpp
+++ b/src/cgwthr.cpp
@@ -60,6 +60,7 @@ void FC cgWthrClean( 		// cgwthr overall init/cleanup routine
 
 }		// cgWthrClean
 //---------------------------------------------------------------------------
+#if defined( TRAPMUNGEDWFNAME)
 bool WfCheck(const char* where)
 {
 	extern TOPRAT Topi;
@@ -80,6 +81,7 @@ bool WfCheck(const char* wfName, const char* where)
 	return bRet;
 
 }	// ::WfCheck
+#endif
 //---------------------------------------------------------------------------
 RC TOPRAT::tp_WthrBegDay()
 {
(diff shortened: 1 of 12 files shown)


src/cgwthr.cpp
```diff
@@ -60,6 +60,29 @@ void FC cgWthrClean( 		// cgwthr overall init/cleanup routine
 
 }		// cgWthrClean
 //---------------------------------------------------------------------------
+#if defined( TRAPMUNGEDWFNAME)
```
nealkruis, line 63: Maybe add a comment with issue number for future reference.

src/xiopak.cpp
```diff
@@ -577,11 +577,25 @@ int xfWriteable(			// check whether file is writeable
...
+	// 1st try -- just use xfExist
+	//   Does not return -1 for at least some bad paths
+	//   TODO: debug/refine xfExist
+	int xfExistRet = xfExist(fPath);
+
+	return xfExistRet >= 0;
+
+}	// xfIsValidPathSyntax
+#endif
+//------------------------------------------------------------------------
```
nealkruis, line 599: Remove and/or add issue?

nealkruis (approved):

src/xiopak.cpp
chipbarnaby: Added comments and issue.

PR #655, opened 2026-07-02
Support Linux builds

### LINKED ISSUE #618: Address case sensitivity
Linux systems use a case-sensitive file system. Supporting this requires us to remove code that automatically converts file paths to uppercase.

Look for instances of `_strupr`.

nealkruis:
Mostly file name case conventions. @SACriswell, note the new input to preserve the old upper case file name convention.

Fixes #618.

## Update (review follow-up)

Addressed feedback from @chipbarnaby's review:
- Clarified `ucFileName` documentation for the `NO` case, and fixed a stale/inaccurate "Locating Files" doc section found along the way (input/`#include`/weather file search order, and the case-insensitive fallback behavior).
- `xfCaseCorrect()`: now reports real filesystem errors, detects and refuses ambiguous case-insensitive matches (rather than silently picking one), and simplified the match comparison per feedback.
- Removed the remaining Windows-only assumptions causing #618: `Path::add()` was unconditionally uppercasing search paths (breaking lookups on Linux), and several filename-parsing helpers (`strffix`, `pointExt`/`pointNamExt`, plus a check in `cncult2.cpp`) assumed `\` as the only path separator.
- Reverted the non-portable `va_list` default-argument commit per @chipbarnaby's request, so his portable fix can merge cleanly on top.

Manually verified on Linux (Docker) since the ambiguous-match path isn't covered by an automated test: confirmed two files differing only by case correctly produce an "Ambiguous case-insensitive match" error and abort the run rather than guessing, and confirmed a single case-mismatched input+weather file resolves via fallback and produces a report file matching the on-disk case (not the case typed on the command line).

## Update (case-sensitive duplicate file names)

Fixed a bug from @chipbarnaby (also in #665): `RFI::rf_CheckForDupFileName()` and `TOPRAT::tp_CheckOutputFilePath()` used `_stricmp()` unconditionally, so on Linux two report/export files differing only by case were wrongly flagged as duplicates.

- Added `xfSamePath()` (`xiopak.h`/`.cpp`), case-insensitive on Windows/macOS, case-sensitive on Linux.
- Both functions now use it. On Linux, case-only differences warn instead of erroring.
- Reconciled with @chipbarnaby's `ISNANDLE` crash-guard fix on the same lines.
- Added `test/error_handling/duplicate_exportfile.cse` for the still-always-an-error case.

Manually verified on macOS and Linux (Docker): same-case duplicates error on both; case-only differences still error on macOS, warn-and-continue on Linux.


doc/src/docs/input-data/top-members.md
chipbarnaby, line 1: ucFileName documentation should be clarified to say what happens when NO.  I assume that the command line case is retained?

Also, what happens on a case-insensitive platform when the command line input file name has different case than the actual file?  That is,

Input file on disk is named baNana.cse and contains ucFileName = NO

cse is invoked CSE Banana

Is the report file baNana.rep or Banana.rep?

I vote for baNana.rep (= match case of input file)

src/RCDEF/rcdef.cpp
chipbarnaby, line 1: This change is unfortunate, but I see that the nullptr default is non-standard.

This has always worked on MSVC and helps with using messages that do not have any args.

I will try to find a way to make a "null" va_list.

src/xiopak.cpp
```diff
@@ -597,6 +597,39 @@ bool xfIsValidPathSyntax(
...
+{
+	std::error_code ec;
+	filesys::path p( tPath);
+	filesys::path dir = p.has_parent_path() ? p.parent_path() : filesys::path(".");
+	std::string target = p.filename().string();
+	if (target.empty() || !filesys::is_directory( dir, ec))
+		return false;
+	for (const auto& entry : filesys::directory_iterator( dir, ec))
+	{	if (ec)
+			break;
```
chipbarnaby: Error should be reported.

src/xiopak.cpp
```diff
@@ -597,6 +597,39 @@ bool xfIsValidPathSyntax(
...
+		std::string entryName = entry.path().filename().string();
+		if (entryName.size() == target.size()
+			&& std::equal( entryName.begin(), entryName.end(), target.begin(),
+				[]( unsigned char a, unsigned char b) { return tolower( a) == tolower( b); }))
+		{	strcpy( tPath + strlen( tPath) - target.size(), entryName.c_str());
+			return true;
```
chipbarnaby: What about the ambiguous case(s)?

Probably need a "could be this or could be that" message.

src/xiopak.cpp
```diff
@@ -597,6 +597,39 @@ bool xfIsValidPathSyntax(
...
+	filesys::path p( tPath);
+	filesys::path dir = p.has_parent_path() ? p.parent_path() : filesys::path(".");
+	std::string target = p.filename().string();
+	if (target.empty() || !filesys::is_directory( dir, ec))
+		return false;
+	for (const auto& entry : filesys::directory_iterator( dir, ec))
+	{	if (ec)
+			break;
+		std::string entryName = entry.path().filename().string();
+		if (entryName.size() == target.size()
```
chipbarnaby: In my opinion, not worth the code to compare lengths.  The compare loop is generally going to exit quickly in most cases where the filenames differ.

chipbarnaby (changes requested):

src/RCDEF/rcdef.cpp
nealkruis: Will this hold up this PR?

chipbarnaby: The internet found me a way to use a default arg in a standard way.

I could make a branch that just fixes that.  If we merge that into this one, there would be lots of conflicts.

Is it possible to revert just the commit for the nullptr fix?  Then merge my version?

nealkruis: I checked all current call sites — both rcderr() and msgWrite() always call va_start before invoking msgWriteV, so the nullptr default was never actually exercised in practice. The signature change should be safe/complete as-is on our end. Let me know if you'd still like to support a "null va_list" on this branch.

nealkruis: Reverted (11217d8a) and pushed to this branch. Note this restores the original nullptr-default pattern, so the Ubuntu CI job will likely fail again until your fix lands on top — go ahead and push/merge whenever it's ready.

chipbarnaby (approved):

PR #660, opened 2026-07-16
Central hpwh update: Add single-pass units from Nyle, AOSmith, Laars, Watts.

tanaya-mankad:
## Description

Add Central HPWH equipment enums for AOSmith, Nyle, Watts, and Laars.
Add enums for Rheem as placeholders, but don't associate with a usable preset yet (there isn't one).

DHW_MFsizing regression test now contains data from an IceAir multi-pass unit.


chipbarnaby (commented):
Big changes in DHW_MFSIZING.REP.  What is up with that?

tanaya-mankad:
> Big changes in DHW_MFSIZING.REP. What is up with that?

We added a new model to the test, that's all.

nealkruis (approved):

Issue #654, opened 2026-07-01
Make Windows 64 bit the basis for regression tests


nealkruis:
Make new regression test reference report files from a 64 bit build and use them as the basis for CI testing.

PR #661, opened 2026-07-23
Make results from win64 be primary regression reference files

chipbarnaby:
## Description

Addresses issue #654.

Committed a starting set of regression results to test/ref-win64-msvc.  Compared all files to corresponding 32 bit version.  Many infinitesimal errors.  A few small but potentially meaningful differences as documented in issue #662.


nealkruis:
@chipbarnaby, do we want to remove the 32-bit regression reference files so there is less to maintain?

nealkruis (approved):

